PR #6, opened 2015-08-17
V2 bug fix for double stop pod and reclaim non-pod vm

gao-feng:
Signed-off-by: Gao feng <feng@hyper.sh>


gao-feng:
xen doesn't have qmp. So I think it's unnecessary.

gao-feng:
@carmark commit `remove pod status check in stoppod from hyper to runv` updated. please check,thanks!

PR #7, opened 2015-08-18
rename detatch to detach

gao-feng:


gao-feng:
@carmark yes, it's good. but I don't know if golang or something will do spell check for us.

carmark:
@gao-feng I don't think golang will do the spell check for us, we should do it by ourselves.

gao-feng:
ok, that's another patch, please review this one.

carmark:
LGTM

PR #10, opened 2015-08-21
Network related update

gao-feng:
1, refactor network initialization codes, and fix incorrectly call network,InitNetwork in vbox case
2, introduce disable iptables feature


gao-feng:
test-by: @feiskyer , merged.

PR #11, opened 2015-08-21
add GET_POD_IP command to get pod's IP address

carmark:


gao-feng:
hi, what do you think about this patch 'https://github.com/hyperhq/runv/pull/12'? it's another way to implement this feature.

carmark:
@gao-feng Will you use this patch to implement the network interfaces of pod setup?

gao-feng:
with #12, you can get pod ip from mypod.Ip directly,no need to send getpodip request and read ip from response.

carmark:
@gao-feng I think my propose has better scalability.  You may fill in more data into response when the client request more information.

And with your PR, there are no `StartedAt` and `FinishedAt` fields in Pod structure and `ExitCode` in Container structure.

laijs:
LGTM

PR #12, opened 2015-08-21
save the ip address of pod in hypervisor.pod structure

gao-feng:


gao-feng:
#11  merged, close

PR #15, opened 2015-08-24
fix compile error on darwin

gao-feng:
And remove the dead codes


carmark:
It seems that you need to `go fmt` the file `hypervisor/network/network_darwin.go`, others look good to me.

PR #19, opened 2015-08-28
remove the dependence of godep

laijs:
The path Godeps/_workspace/ is added to GOPATH when build.
so godep itself doesn't be required for build,  remove it.

Note1: godep is still needed for the developers who want to
update the Godeps/
Note2: the golang will have offical vendor support
in the near future.


gnawux:
+1

PR #21, opened 2015-09-01
update network_darwin.go network_unsupported.go

laijs:


carmark:
The network on darwin should be unsupported, so shall we delete `network_darwin.go` file and modify `network_unsupported.go` file to make it work on darwin system?

laijs:
It is orthogonal to the pr.

gao-feng:
LGTM

carmark:
@laijs  

OK, LGTM

PR #22, opened 2015-09-01
Runv subcommand

laijs:
as runc's subcommand

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/codegangsta/cli"
+)
+
+const (
+	version = "0.2"
+	usage   = `Open Container Initiative hypervisor-based runtime
+
+runv is a command line client for running applications packaged according to
+the Open Container Format (OCF) and is a compliant implementation of the
+Open Container Initiative specification.  However, due to the difference
+between hypervisors and containers, the following sections of OCF don't
+apply to runV:
+    Namespace
+    Capability
+    Device
+    "linux" and "mount" fields in OCI specs are ignored
+
+The current release of "runV" supports the following hypervisors:
+    KVM (QEMU 2.0 or later)
+    Xen (4.5 or later)
+    VirtualBox (Mac OS X)
+
+After creating a spec for your root filesystem, you can execute a container
+in your shell by running:
+
+    cd /mycontainer
+    runv [ spec-file ]
+
+or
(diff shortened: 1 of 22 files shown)


gao-feng:
why introduce the subcommands ? is there other subcommands except start?

laijs:
Other subcommands will be added.
The goal is to make runv more compatible with the runc as we can.
The user don't need to learn to much when he/her knows runc.

gao-feng:
compatible with runc is good, with this patch, can `runv` successfully run without start subcommand?

laijs:
Yes, it can, the same as runc.
The older command line still work

main.go
```diff
@@ -0,0 +1,94 @@
...
+    "linux" and "mount" fields in OCI specs are ignored
+
+The current release of "runV" supports the following hypervisors:
+    KVM (QEMU 2.0 or later)
+    Xen (4.5 or later)
+    VirtualBox (Mac OS X)
+
+After creating a spec for your root filesystem, you can execute a container
+in your shell by running:
+
```
gao-feng: The above messages looks wired stay in usage...

main.go
```diff
@@ -0,0 +1,94 @@
...
+			Usage: "[WIP] enable debug output for logging",
+		},
+		cli.StringFlag{
+			Name:  "log",
+			Usage: "[WIP] set the log file path where internal debug information is written",
+		},
+		cli.StringFlag{
+			Name:  "root",
+			Value: "/run/oci",
+			Usage: "[WIP] root directory for storage of container state (this should be located in tmpfs)",
```
gao-feng: Add unsupport tag in the usage of these  four flags?

gao-feng:
had better to add some commit logs to show how to use runv.

main.go
laijs: The first line reads it supports OCF, but actually it is not fully support, so the compatible note is required after it.
The hypervisor section is also useful information here.

laijs:
`runv --help` will shows how to use runv

main.go
laijs: They will be supported.

main.go
gao-feng: these messages are already on runv readme. usage is used to show how to use this program, isn't it?

main.go
gao-feng: Then add them back with the implementation?

laijs:
updated, merge it

PR #26, opened 2015-09-14
update runv to support oci specs

gao-feng:
up to oci specs commit `dca1dfdd92405129fba7ba0c91cfade2dae1d615`


laijs:
I don't think we need the `runv spec` command. Even the `runc spec` is not necessary for the future exec-driver.

gao-feng:
The spec `runv spec` generates may provides some different configuration  with other runtimes.

PR #27, opened 2015-09-14
portable shell function definition in case /bin/sh is not bash

justincormack:
autogen.sh does not currently build on Debian/Ubuntu as uses bash specific syntax


laijs:
Thanks

PR #29, opened 2015-09-16
do not link xen libs, dynamic load them instead

gnawux:
then we can build unify binary for all platform, no matter whether
they have xen libs installed.


laijs:
LGTM

PR #33, opened 2015-09-22
Init tty connection before container run

gnawux:
originally we start pod, then attach tty to it. Then we may miss
the beginning of the tty output. Even miss all the output if the
progress exit at once.

this patch include the hyper part and runv part, to enable
establishing the tty connection before starting pod. The result
will be like:

    gnawux@sonic$ ./hyper run busybox ps aux
    PID   USER     COMMAND
        1 root     /init
        2 root     ps aux
    Successfully attached to pod(pod-xhZAmXwXCV)
    End of CmdExec(), Waiting for hijack to finish.
    POD id is pod-xhZAmXwXCV

This is required by hyper pr 84

### DIFF
--- a/hypervisor/context.go
+++ b/hypervisor/context.go
@@ -50,6 +50,7 @@ type VmContext struct {
 
 	ptys        *pseudoTtys
 	ttySessions map[string]uint64
+	pendingTtys []*AttachCommand
 
 	// Specification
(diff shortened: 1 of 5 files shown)


gnawux:
code re-formatted 

pending for merge

PR #35, opened 2015-09-22
remove default action of runv

laijs:


laijs:
make it the same as new runc.

PR #36, opened 2015-09-24
Commandline

laijs:


gnawux:
I don't like the filename `spec.go`, it looks like some spec defined in it

laijs:
the sub command name is spec. runc/spec.go provides the subcommand and and the loadSpec(). I think it is ok. the specification has its own package.

gao-feng:
LGTM

PR #39, opened 2015-09-26
refactor the vm chan, implement a fanout

gnawux:
now use the fanout to manage the response channel with the observer pattern

I implement a general fanout with chan, however, golang does not support general programming. wtf


laijs:
LGTM

PR #42, opened 2015-09-28
when fanout close called multiple times, might lead to invalid memory access

gnawux:
check nil inside Close method


gao-feng:
lgtm

PR #45, opened 2015-10-04
add support of separate stderr output of container

gnawux:
this patch add a pipe besides the pty for container, and the stderr will be
sent from the stderr session if the session provided in spec.

this intend to support log the stdout and stderr separately.


gnawux:
rebased on #46

gnawux:
this depends on https://github.com/hyperhq/hyperstart/pull/12

gao-feng:
code LGTM

PR #46, opened 2015-10-06
Rename hypervisor.Pod to hypervisor.PodStatus

gnawux:
Too many things called Pod


gnawux:
just a rename, I will merge it at once

PR #48, opened 2015-10-11
only create stderr seq for those non-interactive pod

gnawux:
then for shells, it can work in interactive mode. and for daemons, they can output stdout and stderr to different log streams.


gao-feng:
LGTM

PR #50, opened 2015-10-12
Support runv exec

gao-feng:
`runv -id 'xx' start ` will create runv container whose id is 'xx'
`runv -id 'xx' exec process.json` will execute program in container 'xx'

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -1,0 +1,124 @@
+package main
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"path"
+
+	"github.com/codegangsta/cli"
+	"github.com/hyperhq/runv/hypervisor"
+	"github.com/hyperhq/runv/lib/term"
+	"github.com/opencontainers/specs"
+)
+
+var execCommand = cli.Command{
+	Name:  "exec",
+	Usage: "exec a new program in runv container",
+	Action: func(context *cli.Context) {
+		config, err := loadProcessConfig(context.Args().First())
+		if err != nil {
+			fmt.Errorf("load process config failed %v\n", err)
+			os.Exit(-1)
+		}
+		if os.Geteuid() != 0 {
+			fmt.Errorf("runv should be run as root\n")
+			os.Exit(-1)
+		}
+		status, err := execProcess(context, config)
+		if err != nil {
+			fmt.Errorf("exec failed: %v", err)
+		}
+		os.Exit(status)
+	},
+}
+
+func execProcess(context *cli.Context, config *specs.Process) (int, error) {
+	podId := context.GlobalString("id")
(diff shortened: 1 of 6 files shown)


runv.go
```diff
@@ -82,6 +86,73 @@ func getTtySize(outFd uintptr, isTerminalOut bool) (int, int) {
...
+
+	pwd, err := filepath.Abs(".")
+	if err != nil {
+		fmt.Printf("%s\n", err.Error())
+		return nil, err
+	}
+
+	state := specs.State{
+		Version: specs.Version,
+		ID:      vmId,
```
laijs: the id should be the containerid which is repsented by podid in this context.

runv.go
```diff
@@ -82,6 +86,73 @@ func getTtySize(outFd uintptr, isTerminalOut bool) (int, int) {
...
+	if err != nil {
+		fmt.Printf("%s\n", err.Error())
+		return nil, err
+	}
+
+	defer func() {
+		if err != nil {
+			os.RemoveAll(podPath)
+		}
+	}()
```
laijs: here is the end of the function, it is awkward to a defer

hypervisor/init_comm.go
```diff
@@ -11,8 +11,8 @@ import (
 
 // Message
 type DecodedMessage struct {
-	code    uint32
-	message []byte
+	Code    uint32
+	Message []byte
 }
```
laijs: use a separate commit to do the rename.

exec.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+		// Discard errors due to pipe interruption
+		sendStdin <- nil
+	}()
+
+	if err := <-receiveStdout; err != nil {
+		return -1, err
+	}
+	sendStdin <- nil
```
laijs: I can't understand this line

runv.go
gao-feng: I whould like to store more information in state.json, the podid can be got by directory name. I also consider to store containerid not vmid in state.json. Let's decide it later. BTW, the state.json of runc is not compatible with ocf right now.

hypervisor/init_comm.go
gao-feng: There is no big difference to seperate one commit to two,
exec support depends on this change, so they are in the same commit.
there is nothing wrong.

exec.go
gao-feng: Me too, just copy it from hyper/client/hijack.go#hijack, with my test, this line is needed

runv.go
gao-feng: I whould like to store more information in state.json, the podid can be got by directory name. I also consider to store containerid not vmid in state.json. Let's decide it later. BTW, the state.json of runc is not compatible with ocf right now.

laijs:
LGTM

PR #52, opened 2015-10-13
avoid the stdio to be closed

laijs:
if the stdio is closed, the output from runv itself is lost
and the 'term.RestoreTerminal(inFd, oldState)' fails.

so we use pipes to prevent the stdio from closing.


gao-feng:
LGTM

PR #54, opened 2015-10-14
cleanup

gao-feng:
1, introduce startvm for runv start
2, introduce new response type E_POD_IP


runv.go
```diff
@@ -318,31 +352,22 @@ func startVContainer(context *cli.Context) {
...
+	defer func() {
+		Response = vm.StopPod(mypod, "yes")
+
+		if Response.Data == nil {
+			fmt.Printf("StopPod fail: QEMU response data is nil\n")
+			return
+		}
+		fmt.Printf("result: code %d %s\n", Response.Code, Response.Cause)
+	}()
+
```
carmark: If the error is encountered before the line of `339`, the `defer` function may panic since there is no `mypod` object, right?

gao-feng: http://blog.golang.org/defer-panic-and-recover says:
>A defer statement pushes a function call onto a list. The list of saved calls is executed after the surrounding function returns.
I think this means, defer function is pushed to the list only after program runs to 361,  so if err occurs before 339, this defer function is not on the list, it will not be called.

carmark: OK, my misunderstand.

code LGTM

PR #56, opened 2015-10-19
support configure sysctl for container

gao-feng:
based on https://github.com/hyperhq/hyperstart/pull/17
```
{
        "id": "test-container-sysctl",
        "containers" : [{
            "name": "ubuntu",
            "image": "ubuntu:latest",
            "workdir": "/",
            "command": ["/bin/bash"],
            "sysctl":{"net.ipv4.ip_forward": "1", "net.core.somaxconn": "256"}
        }],
        "resource": {
            "vcpu": 1,
            "memory": 512
        },
        "tty": true
}
```

### DIFF
--- a/hypervisor/context.go
+++ b/hypervisor/context.go
@@ -299,6 +299,7 @@ func (ctx *VmContext) InitDeviceContext(spec *pod.UserPod, wg *sync.WaitGroup,
 		ctx.initContainerInfo(i, &containers[i], &container)
 		ctx.setContainerInfo(i, &containers[i], cInfo[i])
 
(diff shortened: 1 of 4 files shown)


hypervisor/pod/ocf.go
```diff
@@ -7,11 +7,12 @@ import (
 	"strings"
 )
 
-func OCFSpec2Pod(s specs.Spec, memory int) *UserPod {
+func OCFSpec2Pod(s specs.Spec, memory int, sysctl map[string]string) *UserPod {
 	container := UserContainer{
```
laijs: it should use the spec.RuntimeSpec

PR #59, opened 2015-10-20
Runv exec resizable tty window

laijs:
add respond tag
resizable tty window from unix sock

and these are preparing patches for later sharing-ns `runv start`

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -1,0 +1,98 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+	"os"
+	"os/signal"
+	"path"
+	"syscall"
+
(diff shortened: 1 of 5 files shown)


tty.go
```diff
@@ -0,0 +1,97 @@
...
+		return
+	}
+
+	height, width := getTtySize(tty.termFd)
+	if tty.vm != nil {
+		tty.vm.Tty(tty.tag, height, width)
+		return
+	}
+
+	conn, err := net.Dial("unix", path.Join(tty.containerDir, "runv.sock"))
```
gao-feng: resizeTty can be called many time before runv exit, had better to close conn after use it.

laijs: updated

PR #62, opened 2015-10-22
Cleanup

laijs:
rename podId to container

The podId, podPath are for container when old simple container
implementation, we will enable multiple containers. So the renaming
is required.

podId -> container
podPath -> stateDir
containerDir -> stateDir
save container ID in state file


gao-feng:
LGTM

PR #63, opened 2015-10-22
rename ocf_linux.go to ocf-linux.go and the function names

laijs:
And remove the other related files.
We only support linux container.
Bot host (linux & osx) are supported and are the same functional
for the linux container.


gao-feng:
LGTM

PR #67, opened 2015-10-29
Cleanup

laijs:
simplify hooks executing

move hypervisor/utils.go to lib/utils/
   
fix the warning for daemon.go


carmark:
code LGTM

(this comment is not related with the fix)
It may print the error twice in `runv.go` when returning an error from a function, such as `preparePod`, `saveState` and so on.

laijs:
The runv.go is/will be changing quickly. The error handling may also be changed. thx.

PR #72, opened 2015-11-05
get right response of request

gao-feng:
1,the vm state machine doesn't handle unexpected event(request), there will be no response to this request, this will cause the caller blocked to waiting for response.
first patch fixs this by sending unexpected response to request if request needs

2, right now we don't check if exec event is handled correctly. the second patch add this check, so we don't try to read stream callback if exec event failed. And other events get response through response chan, this patch makes exec event consist with other events. This patch also introduces Replyto field to
Vmresponse, with this field, request can check if the response is what it expected.

3, Fix possible getting incorrect response for read/write file event

TODO: Fix possible getting incorrect response for RunPod,StopPod events

### DIFF
--- a/hypervisor/events.go
+++ b/hypervisor/events.go
@@ -85,14 +85,11 @@ type AttachCommand struct {
 }
 
 type CommandAck struct {
-	reply uint32
+	reply *DecodedMessage
 	msg   []byte
 }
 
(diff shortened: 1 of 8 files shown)


laijs:
LGTM

PR #73, opened 2015-11-05
New container

laijs:
new container function(inserting container)
split pod&container creation in runv.

These are the hard part.
cleanup&runv-daemon&shared-ns-runv-start will be implemented based on these soon


gao-feng:
LGTM

PR #74, opened 2015-11-09
Runv ns daemon

laijs:


laijs:
all changes are included in #75, close it.

PR #75, opened 2015-11-11
Shared namespaces multiple containers support

laijs:
The runv supports pod-style shared namespaces currently.
(More fine grain shared namespaces style (docker/runc style) is under implementation)

Pod-style shared namespaces:
 * if two containers share at least one type of namespace, they share all kinds of namespaces except the mount namespace
 * mount namespace can't be shared, each container has its own mount namespace

 Implementation detail:
 * Shared namespaces is configured in LinuxRuntimeSpec.Linux.Namespaces, the namespace Path should be existing container name.
 * In runv, shared namespaces multiple containers are located in the same VM which is managed by a runv-daemon.
 * Any running container can exit in any arbitrary order, the runv-daemon and the VM are existed until the last container of the VM is existed

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -1,17 +1,14 @@
 package main
 
 import (
-	"bufio"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net"
 	"os"
 	"path"
 
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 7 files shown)


laijs:
Discussed, merged.

PR #78, opened 2015-11-17
pass Abs path of kernel/initrd to runv daemon

gao-feng:


laijs:
need to change the default behavior a little. but we merge it at first, and then I change them later.

PR #89, opened 2015-12-03
Update glog

laijs:


carmark:
Do not merge this patch before modifying the `hyper` code.  Since `hyper` will use `runv/lib/glog`, we have to make sure `hyper` can build successfully.

PR #93, opened 2015-12-17
(do NOT merge) Enable lazy launch for qemu driver

laijs:
Current persistent memory is not support hotplug,
so we add the lazy launch functionality to the qemu driver
and perpare for adding dax functionality (via persistent memory)
in future.
(the lazy launch functionality = the vm is not launched until startpod)

All lazy functions are moved to the DriverContext interface,
and the LazyDriverContext is removed.

The interface HypervisorDriver adds AsyncDiskBoot() and AsyncNicBoot(),
and when either one of it returns false, the driver should use
the lazy launch functionality to lunch the vm.

When persistent memory is supported, the qemu driver only brings disks
to qemu argument when boot. The Nics are still async boot.

### DIFF
--- a/hypervisor/vm.go
+++ b/hypervisor/vm.go
@@ -54,10 +54,10 @@ func (vm *Vm) Launch(b *BootConfig) (err error) {
 		Status   = make(chan *types.VmResponse, 128)
 	)
 
-	if vm.Lazy {
-		go LazyVmLoop(vm.Id, PodEvent, Status, b, vm.Keep)
-	} else {
+	if HDriver.AsyncDiskBoot() && HDriver.AsyncNicBoot() {
 		go VmLoop(vm.Id, PodEvent, Status, b, vm.Keep)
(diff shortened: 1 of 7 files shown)


hypervisor/vm.go
```diff
@@ -54,10 +54,10 @@ func (vm *Vm) Launch(b *BootConfig) (err error) {
 		Status   = make(chan *types.VmResponse, 128)
 	)
 
-	if vm.Lazy {
-		go LazyVmLoop(vm.Id, PodEvent, Status, b, vm.Keep)
-	} else {
+	if HDriver.AsyncDiskBoot() && HDriver.AsyncNicBoot() {
```
gao-feng: Can we remove SupportLazyMode?

laijs: yes, it can be removed, but after the hyper.git site removes the reference to it.

PR #97, opened 2016-01-05
move the code of probing the fallback drivers into the package driver…

laijs:
…loader

### DIFF
--- a/driverloader/driverloader_linux.go
+++ b/driverloader/driverloader_linux.go
@@ -2,38 +2,43 @@ package driverloader
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/hyperhq/runv/hypervisor"
 	"github.com/hyperhq/runv/hypervisor/libvirt"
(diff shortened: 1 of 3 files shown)


driverloader/driverloader_linux.go
```diff
@@ -2,38 +2,43 @@ package driverloader
...
+		if _, err := os.Stat("/dev/kvm"); os.IsNotExist(err) {
+			return nil, fmt.Errorf("Driver %s is unavailable\n", driver)
 		}
-	case "xen":
+		qd := qemu.InitDriver()
+		if qd != nil {
+			fmt.Printf("%s Driver Loaded\n", driver)
+			return qd, nil
+		}
+	case "xen", "":
```
gao-feng: Does this mean we use xen driver if no hypervisor specified?

laijs: only if the xen driver is available. otherwise it will use qemu.
and qemu driver uses kvm if kvm is available.

gao-feng:
LGTM

PR #103, opened 2016-01-11
shutdown libvirt vm with destroy

gao-feng:
shutdownflags doesn't shutdown vm actually, use destroyflags
to make sure shutdown vm successfully.


feiskyer:
LGTM

PR #107, opened 2016-01-18
Fix volume filename

### LINKED ISSUE #108: Rbd volume not working for qemu driver
Monitors and secrets not setted up for rbd volume

feiskyer:
`setVolumeInfo` should not override volume name setted up at `initVolumeMap`

Fix #108

### DIFF
--- a/hypervisor/devicemap.go
+++ b/hypervisor/devicemap.go
@@ -230,7 +230,10 @@ func (ctx *VmContext) setVolumeInfo(info *VolumeInfo) {
 		return
 	}
 
-	vol.info.filename = info.Filepath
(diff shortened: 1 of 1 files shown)


gnawux:
I am afraid this will affect other volume drivers. It looks this changes the behavior of the method.

feiskyer:
@Gnep Why changes the behavior?

gnawux:
the original line is an unconditional assignment, and in this patch, it only assign when the `len()` is zero. I am not sure about the detailed, but you should show it does not affect or the original is wrong for all drivers before merge.

feiskyer:
See #110

PR #110, opened 2016-01-18
Fix rbd volume for libvirt driver

### LINKED ISSUE #109: Rbd volume not working for libvirt driver
Currently, runv generates spec for rbd volue as follows:

```xml
<disk type="file" device="disk"><driver><Type>rbd</Type></driver><source file="rbd:test/rbd-image"></source><target dev="sdc" bus="scsi"></target><address type="drive" controller="0" bus="0" unit="2"></address></disk>
```

While the correct rbd volume spec for rbd volume is 

```xml
<disk type='network' device='disk'>
        <source protocol='rbd' name='libvirt-pool/new-libvirt-image'>
                <host name='{monitor-host}' port='6789'/>
        </source>
        <auth username='libvirt'>
                <secret type='ceph' uuid='9ec59067-fdbc-a6c0-03ff-df165c0587b8'/>
        </auth>
        <target dev='vda' bus='virtio'/>
</disk>
```

Ref http://docs.ceph.com/docs/master/rbd/libvirt/

feiskyer:
Fix #109 #108

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -9,8 +9,8 @@
 		},
 		{
 			"ImportPath": "github.com/alexzorin/libvirt-go",
-			"Comment": "v2.12.0-2-g3793956",
-			"Rev": "379395637960a7e11936e61344ab761aedcdb44e"
+			"Comment": "v2.12.0-2-g057f70a",
+			"Rev": "057f70a36a6b5d4ea19be345c2369e27dd022e01"
 		},
 		{
 			"ImportPath": "github.com/codegangsta/cli",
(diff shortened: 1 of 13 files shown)


gnawux:
could the `VolumeOptions` be a `map[string]string`? may be useful for other types of storage

feiskyer:
It could, But I think struct is more clear than map.

gnawux:
define a static structure is helpful for understand and testing, but this struct is dedicated for rbd, not a general volume options.

feiskyer:
@gnawux  See https://github.com/hyperhq/runv/blob/master/hypervisor/pod/pod.go#L68. Although those params are only used in rbd volume, they could also be used for other new volumes.

laijs:
`AddDisk(ctx *hypervisor.VmContext, source string, options map[string] []string)`

options: map string to vector of strings. (like url.Values)

feiskyer:
@gnawux @laijs Refactored to map[string]string

PR #117, opened 2016-02-03
add dedicate hostname field in pod

gnawux:
this is for hyperhq/hyper#200, provide a hostname field in pod spec

the test case for hyper will be uploaded after this is merged


feiskyer:
LGTM

PR #128, opened 2016-03-03
Factor vm hub channel

laijs:
rename vm.VmChan to vm.Hub
and
remove vm.GetRequestChan() and vm.ReleaseRequestChan()

so we don't need to copy this dummy code every time for every new api


gao-feng:
LGTM

PR #134, opened 2016-03-04
don't recevice stdin if container/exec haven't started

gao-feng:
Since the container/exec may haven't started yet, if hyperstart
received the unknown session' tty data, it will send eof message
to notify hyperd to close the connection.

You can reproduce this problem by press enter button all the time
after execute `hyper run -t image cmd`

Fix this problem by don't receive stdio input data before hyperd
get the ack of startpod/newcontainer/exec.


laijs:
The pty code is refactoring, the fix will be pending until the refactor is done.

PR #135, opened 2016-03-05
Ptys refactor part2

laijs:
remove the wrong linerTransformer

the second path simplifies Attachments allocation, all are allocated
in ptyConnect(), and it  doesn't change any functionality.


gao-feng:
LGTM

PR #137, opened 2016-03-08
change VmContainer's stdio's protocal and send tty into hyperstart

laijs:
VmContainer.Tty: allocate tty or not
VmContainer.Stdio: Stdio sequence number
VmContainer.Stderr: Stderr sequence number if stderr is not share with
stdout (currently used when VmContainer.Tty is false)


gnawux:
restarted the test, and passed now.

PR #138, opened 2016-03-16
use per-container tty conf in pod spec

gnawux:
- global tty conf in pod spec is now deprecated
- currently container tty option will be (pod.tty || container.tty)


laijs:
LGTM

PR #140, opened 2016-03-18
Support flying vm without hyperd manage

gao-feng:
To support flying vm, hyper will create libvirt domain without AUTODESTROY flag to make sure domain alive when hyperd exited. 
And setup started flag for tty attachment of container to allow client to input data.

### DIFF
--- a/hypervisor/tty.go
+++ b/hypervisor/tty.go
@@ -192,11 +192,16 @@ func waitPts(ctx *VmContext) {
 }
 
 func newAttachmentsWithTty(persist, isTty bool, tty *TtyIO) *ttyAttachments {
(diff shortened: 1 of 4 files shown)


hypervisor/tty.go
```diff
@@ -323,13 +327,20 @@ func (pts *pseudoTtys) ptyConnect(persist, isTty bool, session uint64, tty *TtyI
 	pts.lock.Unlock()
 }
 
-func (pts *pseudoTtys) startStdin(session uint64) {
+func (pts *pseudoTtys) startStdin(isTty bool, session uint64) {
```
gnawux: looks parameters in wrong order... XD

PR #141, opened 2016-03-19
send stdin eof to hyperstart when receive eof from client

laijs:
when the last attached stdin is eof, we also send eof
of the stdin to the container and we don't detach the
io, becasue we need send the last output to the client.

this enable pipes functionalities:
echo 12345678 | hyper run ubuntu cat
echo ls | hyper run ubuntu


gao-feng:
LGTM

PR #144, opened 2016-03-27
pid unmarshal from persist info should be float64 type

Crazykev:
With this bug fixed, If `hyperd` is forced to be killed and start again, reconnecting to the running pod feature should function normally. 


laijs:
Good, Thanks!

PR #146, opened 2016-03-28
libvirt: do not access domain if create failed

gao-feng:
fix panic when create domain failed.


bergwolf:
LGTM!

While it is robust to defend against crazy callers, I think we should also fix anyone that calls libvirt functions after failing to set the domain context, no?

gao-feng:
You are right, it's a little complex, I need more time to consider it.

PR #152, opened 2016-04-02
revert 208e30f3 and add balloon back

laijs:
208e30f3 is workaroud for the bug of the newest qemu at that time.
but the stable qemu(2.4.1) doesn't have such bug.

revert it. sigh...


bergwolf:
LGTM!

PR #158, opened 2016-04-04
set the default detachkeys to "ctrl+p,ctrl+q"

laijs:


laijs:
https://github.com/docker/docker/issues/21769

PR #159, opened 2016-04-05
Use only commands/envs in container info

gnawux:
- the spec info is written by user
- the container info is comes from inspect, which is merged the user config and image config during container creation

Then, we do not need to merge them again.


gnawux:
github announce recovered, let's trigger the test again

laijs:
LGTM

PR #160, opened 2016-04-06
introduce internal volume for hyper add volume

gao-feng:
this give us a way to know if the volume is added by user
or by hyper.


laijs:
the field should be add to VolumeInfo instead.

PR #166, opened 2016-04-08
vm.KillContainer is blocked as vm_states does not reply to the command

gnawux:


laijs:
LGTM

PR #167, opened 2016-04-09
setup pod handler after pod is well associated

gnawux:
In the previous implementation, the associate and handler are two parallel
goroutine. However, the handler goroutine need vmcontext.client to be
precent, which is assigned in the associate goroutine.

In this patch, we associate first, and only let the main loop to be a goroutine.
Then, when the handler start, it must have the vmcontext.client channel ready.

This could guarantee the stop/kill/restart operation still available after re-associate.


gao-feng:
lgtm

PR #175, opened 2016-04-13
Print runv-ns-daemon log via glog

WeiZhang555:
runv-ns-daemon will run under daemon mode and print everything to
/dev/null, which will lost useful debug information.
This commit will allow printing all debug message via glog instead of
Printf, keep the useful log information into log files.

PS: I'm not that familiar to Runv, so if it's more suitable to use `glog.V(3)` in some places instead of `V(1)`, feel free to comment. :smile: 

### DIFF
--- a/hypervisor/pod/ocf-linux.go
+++ b/hypervisor/pod/ocf-linux.go
@@ -1,9 +1,9 @@
 package pod
 
 import (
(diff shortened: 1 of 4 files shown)


laijs:
LGTM

PR #180, opened 2016-04-14
Add runv list command

WeiZhang555:
Add runv list command to show running containers

Final look:
```
$ sudo ./runv list
ID             PID         STATUS      BUNDLE              CREATED
busybox-runv   -1          running     /tmp/busybox-runv   2016-04-14T11:30:36.328931424+08:00
```

### DIFF
--- a/list.go
+++ b/list.go
@@ -1,0 +1,139 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"text/tabwriter"
+	"time"
+
+	"github.com/codegangsta/cli"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+const formatOptions = `table or json`
+
+// containerState represents the platform agnostic pieces relating to a
+// running container's status and state
+type containerState struct {
+	// ID is the container ID
+	ID string `json:"id"`
+	// InitProcessPid is the init process id in the parent namespace
+	InitProcessPid int `json:"pid"`
+	// Status is the current status of the container, running, paused, ...
+	Status string `json:"status"`
+	// Bundle is the path on the filesystem to the bundle
+	Bundle string `json:"bundle"`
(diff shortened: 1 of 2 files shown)


list.go
```diff
@@ -0,0 +1,146 @@
...
+			Usage: `select one of: ` + formatOptions + `.
+
+The default format is table.  The following will output the list of containers
+in json format:
+
+  # runv list -f json`,
+		},
+	},
+	Action: func(context *cli.Context) {
+		flag.Parse()
```
laijs: we don't use flag.Parse() on normal commands.

list.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	var s []containerState
+	for _, item := range list {
+		if item.IsDir() {
+			stateFile := filepath.Join(absRoot, item.Name(), "state.json")
+			fi, err := os.Stat(stateFile)
+			if err != nil && !os.IsNotExist(err) {
+				glog.Errorf("Can't load state file for container %s: %s", item.Name(), err)
+				continue
```
laijs: `runv list` is running foreground, it uses fmt.Printf() to output debug/error message.
only the daemon uses the glog.

WeiZhang555: Right.
I'll change.

list.go
WeiZhang555: This is for removing the complaint of glog `ERROR: logging before flag.Parse`. 
As we shouldn't use glog here, this can be removed.
Will do it.

list.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	list, err := ioutil.ReadDir(absRoot)
+	if err != nil {
+		return nil, err
+	}
+
+	var s []containerState
+	for _, item := range list {
+		if item.IsDir() {
+			stateFile := filepath.Join(absRoot, item.Name(), "state.json")
```
laijs: s/"state.json"/stateJson/

WeiZhang555:
Updated.

PR #184, opened 2016-04-18
Add runv state command

WeiZhang555:
Now we can use state command to check one container's state and status


laijs:
LGTM

PR #188, opened 2016-04-19
Fix finish time for stopped Pod

feiskyer:
Should set finishedTime for Pod while its status changed to S_POD_FAILED

### DIFF
--- a/hypervisor/vm.go
+++ b/hypervisor/vm.go
@@ -322,6 +322,7 @@ func (vm *Vm) StopPod(mypod *PodStatus, stopVm string) *types.VmResponse {
 	}
 
 	mypod.Status = types.S_POD_FAILED
+	mypod.FinishedAt = time.Now().Format("2006-01-02T15:04:05Z")
 	mypod.SetContainerStatus(types.S_POD_FAILED)
 
 	return Response


hypervisor/vm.go
```diff
@@ -322,6 +322,7 @@ func (vm *Vm) StopPod(mypod *PodStatus, stopVm string) *types.VmResponse {
 	}
 
 	mypod.Status = types.S_POD_FAILED
+	mypod.FinishedAt = time.Now().Format("2006-01-02T15:04:05Z")
```
gao-feng: Pod may finish before stop pod manually

feiskyer: y, FinishedAt is already ok for that case.

gnawux: I think the only case that need set finished time is `stopVm != yes`

feiskyer: Not agree, for `pod stop` operation, stopVm is true. `FinishedAt ` should be updated.

gnawux: Hmm, 

Anyway, you are correct

https://github.com/hyperhq/hyperd/blob/master/daemon/pod.go#L1007

Here should be a bug --

if the vm exited, no matter it is terminated by user or the process finished itself, it should call `PodStatus.SetPodContainerStatus()` instead of `SetContainerStatus`

feiskyer: Updated in https://github.com/hyperhq/hyperd/pull/295.

feiskyer:
I'm closing this because `mypod.FinishedAt` has been updated at https://github.com/hyperhq/hyperd/pull/295, so not needing here.

PR #190, opened 2016-04-19
support qemu lazy mode 

YaoZengzeng:


laijs:
Thanks for contribution
it needs rebase
could you add a little more changelog.

PR #192, opened 2016-04-20
support qemu lazy mode

YaoZengzeng:
I make qemu match LazyDriverContext and add three items in QemuContext.So we can launch Qemu with net and block device mounted .The "interface" and "image" in QemuContext is used to store the net and block device configuration.The "callbacks" is used to send the event that devices is ready, so we can start pod.
  Unfortunately, there seems a little problem, at the very begining, we may receive a tty or init message with impossible big length message.So, I do a little trick to find and  read the entire wrong message, and start reading again.Then every thing is OK.But I dont know if it is relevant with lazy mode?

### DIFF
--- a/hypervisor/init_comm.go
+++ b/hypervisor/init_comm.go
@@ -79,6 +79,11 @@ func ReadVmMessage(conn *net.UnixConn) (*DecodedMessage, error) {
 		res = append(res, buf[:nr]...)
 		read = read + nr
 
+		// if condition satisified, message length may wrong , read again
+		if nr < want && read < needRead {
+			return ReadVmMessage(conn)
+		}
+
 		glog.V(1).Infof("read %d/%d [length = %d]", read, needRead, length)
 
 		if length == 0 && read >= 8 {
--- a/hypervisor/qemu/qemu.go
+++ b/hypervisor/qemu/qemu.go
@@ -27,6 +27,25 @@ type QemuContext struct {
 	qmpSockName string
 	cpus        int
 	process     *os.Process
+	interfaces 	[]lazyQemuNicConfig
+	images 		[]lazyQemuImageConfig
+	callbacks 	[]hypervisor.VmEvent
+}
+
+//qemu lazy nic configuration
+type lazyQemuNicConfig struct {
+	fd 			uint64
+	deviceName 	string
+	macAddr 	string
+	pciAddr		int
+}
+
+//qemu lazy block configuration
+type lazyQemuImageConfig struct {
+	deviceName 	string
+	sourceType 	string
+	format 		string
+	scsiId 		int
 }
 
 func qemuContext(ctx *hypervisor.VmContext) *QemuContext {
@@ -292,7 +311,7 @@ func (qc *QemuContext) Save(ctx *hypervisor.VmContext, path string, result chan<
 }
 
 func (qc *QemuDriver) SupportLazyMode() bool {
-	return false
+	return true
 }
 
 func (qc *QemuContext) arguments(ctx *hypervisor.VmContext) []string {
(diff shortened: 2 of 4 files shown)


hypervisor/init_comm.go
```diff
@@ -79,6 +79,11 @@ func ReadVmMessage(conn *net.UnixConn) (*DecodedMessage, error) {
 		res = append(res, buf[:nr]...)
 		read = read + nr
 
+		// if condition satisified, message length may wrong , read again
+		if nr < want && read < needRead {
+			return ReadVmMessage(conn)
+		}
+
```
laijs: why these code is needed? the loop will continue read the compose the whole message?

YaoZengzeng: Because it sometimes will get a impossible long message as follows:
I0426 14:22:34.827787   26408 init_comm.go:91] data length is 582570580
I0426 14:22:34.827794   26408 init_comm.go:72] trying to read 512 bytes
I0426 14:22:34.827804   26408 init_comm.go:72] trying to read 8 bytes
I0426 14:22:34.905799   26408 init_comm.go:87] read 8/8 [length = 0]
I0426 14:22:34.905833   26408 init_comm.go:91] data length is 1462868
I0426 14:22:34.905839   26408 init_comm.go:72] trying to read 512 bytes
I0426 14:22:34.905849   26408 init_comm.go:72] trying to read 8 bytes
Especially when you stop the hyper daemon unproperly ,restart it and launch the first pod.So in order not to be blocked here, I do this to clarify the bad message and start to read a new message.

hypervisor/qemu/qemu.go
```diff
@@ -354,4 +373,68 @@ func (qc *QemuContext) arguments(ctx *hypervisor.VmContext) []string {
 		"-fsdev", fmt.Sprintf("local,id=virtio9p,path=%s,security_model=none", ctx.ShareDir),
 		"-device", fmt.Sprintf("virtio-9p-pci,fsdev=virtio9p,mount_tag=%s", hypervisor.ShareDirTag),
 	)
+
+	//lazy add nic device
+	for i, info := range qc.interfaces {
+		params = append(params,
+		"-netdev", fmt.Sprintf("tap,fd=%d,id=%s", info.fd, info.deviceName),
```
laijs: qemu is launched via daemon mode, the all the fds will be closed. how these code works?

YaoZengzeng: Yes，you're right, in this branch I forget something and didn't test the nic after the qemu launch.In order to fix this , we should modify the code under path /runv/lib/utils/daemon.go:
	//Close all open file descriptors
//	for (i = 0; i < num; i++) {
//		if (fds[i] == 0 || fds[i] == 1 || fds[i] == 2)
//			continue;
//		printf("\n\nI didnt close any file descriptor\n\n");
//		close(fds[i]);
//	}
just don't close the fds.If want to optimize , we can just leave the nic fd open.

hypervisor/init_comm.go
laijs: could you split it into another pr?

laijs: it it hard to recover the channel is the previous hyper daemon exited improperly.
we need to find a better solution.

YaoZengzeng: That means I delete this tricky part and just pr the main lazy qemu code???And should I add the code that not closing all fds?

laijs: I will add some code to avoid close some the fds.

YaoZengzeng: OK，I'll make a new PR immediately

laijs: Actually, you don't need to create a new pr for qemu lazy, just use this one.
just create a new pr for ReadVmMessage().

PR #194, opened 2016-04-24
allow dot in hostname

gnawux:
this is what docker do.


gnawux:
@laijs updated the hostname format check

PR #195, opened 2016-04-25
Cleanup vm state machine

gao-feng:


hypervisor/devicemap.go
```diff
@@ -361,27 +360,6 @@ func (ctx *VmContext) netdevInserted(info *NetDevInsertedEvent) {
...
-				glog.V(1).Info("need remove image dm file", image.info.Filename)
-				ctx.progress.deleting.blockdevs[name] = true
-				go UmountDMDevice(image.info.Filename, name, ctx.Hub)
-			}
-		}
-	}
-
-	return c.Success
-}
-
```
laijs: why onContainerRemoved() is also removed?

laijs: OK, it is unused currently. but it is possible needed in future.

gao-feng: then add them back if necessary.

gnawux:
do all the volume cleanup code are removed? I wonder if this will 

- leave the dm devices in `/dev` after the pod stopped
- leave overlay/aufs mount dirs after the pod stopped

gao-feng:
yes, hyperd will do the cleanup job. https://github.com/hyperhq/hyperd/pull/303

PR #197, opened 2016-04-26
Adapt to new OCI spec for root

WeiZhang555:
Now docker use absolute bundle root path instead of relative
[$BUNDLE]/rootfs path, and OCI has made relative changes.

Change runv-containerd to adapt this change, or it won't work any more.

See https://github.com/docker/docker/pull/22256 and https://github.com/opencontainers/runtime-spec/pull/394


laijs:
could you check whether it is absolute path or not at first.
for relative root path, we need to join it to the bundle path.

WeiZhang555:
Yep, you are right. My fault.

WeiZhang555:
updated.

PR #199, opened 2016-04-26
support qemu lazy mode

YaoZengzeng:
I make qemu match LazyDriverContext and add three items in QemuContext.So we can launch Qemu with net and block device mounted .The "interface" and "image" in QemuContext is used to store the net and block device configuration.The "callbacks" is used to send the event that devices is ready, after that we can start pod.

### DIFF
--- a/hypervisor/qemu/qemu.go
+++ b/hypervisor/qemu/qemu.go
@@ -27,6 +27,25 @@ type QemuContext struct {
 	qmpSockName string
 	cpus        int
 	process     *os.Process
+	interfaces  []qemuNicConfig
+	images      []qemuImageConfig
+	callbacks   []hypervisor.VmEvent
+}
+
+//qemu nic configuration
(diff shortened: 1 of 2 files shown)


hypervisor/qemu/qemu.go
```diff
@@ -27,6 +27,25 @@ type QemuContext struct {
 	qmpSockName string
 	cpus        int
 	process     *os.Process
+	interfaces	[]lazyQemuNicConfig
+	images		[]lazyQemuImageConfig
+	callbacks	[]hypervisor.VmEvent
+}
+
+//qemu lazy nic configuration 
+type lazyQemuNicConfig struct {
```
laijs: could you remove the lazy prefix, use qemuNicConfig.
the "lazy" will be rename.

laijs:
gofmt needed

PR #201, opened 2016-04-27
fix possible race in fanout

gao-feng:

### DIFF
--- a/hypervisor/fanout.go
+++ b/hypervisor/fanout.go
@@ -11,10 +11,9 @@ type Fanout struct {
 	upstream chan *types.VmResponse
 	clients  []chan *types.VmResponse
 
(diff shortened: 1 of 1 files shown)


hypervisor/fanout.go
```diff
@@ -62,12 +60,11 @@ func (fo *Fanout) Release(client chan *types.VmResponse) error {
 	for _, c := range fo.clients {
 		if c != client {
 			remains = append(remains, c)
-		} else {
-			fo.unregistered = append(fo.unregistered, c)
+			continue
 		}
+		close(client)
```
laijs: it closes remained channel.
please move it into the else branch and remove the continue

gao-feng: Sorry, I cannot get your point.

PR #204, opened 2016-05-03
Add volume support

WeiZhang555:
Currently runv don't support docker volumes(OCI Mounts fields), this
commit will add volume support, so that we can bind mount host dir or make a
volume dir in container.


supervisor/container.go
```diff
@@ -252,3 +263,130 @@ func (c *Container) reap() {
...
+		}
+		if err := mount.Mount(m.Source, dest, m.Type, strings.Join(m.Options, ",")); err != nil {
+			return err
+		}
+		if stat != nil {
+			if err = os.Chmod(dest, stat.Mode()); err != nil {
+				return err
+			}
+		}
+		return nil
```
laijs: proc sysfs mqueue tmpfs MUST pass into hyperstart and let hyperstart mount it.
hyperstart haven't has such abi, so they should be skipped.
Could you add a issue for such purpose?

supervisor/container.go
```diff
@@ -8,8 +8,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"time"
 
+	"github.com/docker/docker/pkg/mount"
+	"github.com/docker/docker/pkg/symlink"
```
laijs: please check the travis error

supervisor/container.go:14:2: cannot find package "github.com/docker/docker/pkg/mount"

supervisor/container.go
WeiZhang555: OK. I'll skip it.

supervisor/container.go
```diff
@@ -252,3 +263,130 @@ func (c *Container) reap() {
...
+// checkMountDestination checks to ensure that the mount destination is not over the top of /proc.
+// dest is required to be an abs path and have any symlinks resolved before calling this function.
+func checkMountDestination(rootfs, dest string) error {
+	invalidDestinations := []string{
+		"/proc",
+	}
+	// White list, it should be sub directories of invalid destinations
+	validDestinations := []string{
+		// These entries can be bind mounted by files emulated by fuse,
+		// so commands like top, free displays stats in container.
```
gao-feng: seams like runv doesn't need to emulate these files.

WeiZhang555: It looks like that mount points in white list can be used by `fuse` for better isolation. I'm not sure how can runv make use of this, but maybe we can leave it here. 
It wont harm anyone, right?  If not, I can move mount points in white list(validDestinations ) into black list(invalidDestinations)

WeiZhang555:
Updated. Test is green now. :smile:

laijs:
related: https://github.com/hyperhq/hyperd/issues/285

PR #205, opened 2016-05-04
don't close the client event listeners until the context is closed

laijs:
ctx.Close() will close the upstream of the clients, so event listeners
(and Fanout) are closed in consequence.

So that all the event listeners can listen all the event until end.


gao-feng:
LGTM

PR #211, opened 2016-05-11
Disable libnetwork prestart hook

WeiZhang555:
Disable libnetwork-setkey prestart hook right now because actually we
don't support CNM. Before this, if we don't specify `--net none` in docker
command line, something weird will happen, you can see container running
with ps, but you can exec/stop/rm that container, this will influence
user experience largely.


gao-feng:
please hold on, I'm working on  CNM support.

WeiZhang555:
Oh, that's great! 
Supporting CNM is definitely better choice.

laijs:
close it, since network namespace manipulation proxy is merged. #218

PR #216, opened 2016-05-13
mv the structs VmXXX and related method in hypervisor/pod.go to hyperstart/api/json/spec.go

YaoZengzeng:


hypervisor/pod.go
```diff
@@ -123,7 +124,7 @@ type VmPod struct {
 	Dns        []string       `json:"dns,omitempty"`
 	Routes     []VmRoute      `json:"routes,omitempty"`
 	ShareDir   string         `json:"shareDir"`
-}
+}*/
```
laijs: remove them directly

hyperstart/api/json/vm_spec.go
```diff
@@ -0,0 +1,102 @@
+package json
+
+type VmVolumeDescriptor struct {
```
laijs: remove all the Vm prefix.
s/vm_spec.go/spec.go/

hypervisor/pod.go
YaoZengzeng: Oh， I forget it！

hyperstart/api/json/vm_spec.go
```diff
@@ -0,0 +1,102 @@
...
+	Id            string               `json:"id"`
+	Rootfs        string               `json:"rootfs"`
+	Fstype        string               `json:"fstype,omitempty"`
+	Image         string               `json:"image"`
+	Addr          string               `json:"addr,omitempty"`
+	Volumes       []VmVolumeDescriptor `json:"volumes,omitempty"`
+	Fsmap         []VmFsmapDescriptor  `json:"fsmap,omitempty"`
+	Sysctl        map[string]string    `json:"sysctl,omitempty"`
+	Process       VmProcess            `json:"process"`
+	Entrypoint    []string             `json:"-"`
```
laijs: remove Entrypoint directly.
it is referenced in one place only. it is assigned without any other reference. so it can be removed.
correct me if I was wrong.

YaoZengzeng: It seems OK,but without "Entrypoint" item, the starting of container wont meet any problem?Or the hyperstart just process the nonexistent "Entrypoint" item by default?

laijs:
please fix the commit title
avoid to use master branch when create pr.

PR #218, opened 2016-05-16
runv-containerd: allow to start docker with network

gao-feng:
based on https://github.com/hyperhq/hyperstart/pull/95


hypervisor/devicemap.go
```diff
@@ -327,8 +327,24 @@ func (ctx *VmContext) blockdevInserted(info *BlockdevInsertedEvent) {
...
+	}
+	g := &GuestNicInfo{
+		Device:  info.DeviceName,
+		Ipaddr:  info.IpAddr,
+		Index:   info.Index,
+		Busaddr: info.PCIAddr,
+	}
+
+	ctx.DCtx.AddNic(ctx, h, g)
 }
```
laijs: lazyAllocateNetworks() also calls interfaceCreated() and LazyAddNic().
I guess it will be bug. when lazyAllocateNetworks() is called.

gao-feng: yes, will fix this

supervisor/hyperpod.go
```diff
@@ -90,13 +237,21 @@ func createHyperPod(f factory.Factory, spec *specs.Spec) (*HyperPod, error) {
...
 		podStatus:  podStatus,
 		vm:         vm,
 		Containers: make(map[string]*Container),
 		Processes:  make(map[string]*Process),
-	}, nil
+	}
+
+	// create Listener process running in its own netns
+	if err = hp.startNsListener(); err != nil {
+		glog.V(1).Infof("%s\n", err.Error())
```
laijs: go hp.reap()

laijs: when vm.StartPod() fails. the original code doesn't kill the vm. it is a bug blamed to @laijs.
you can also add it.

supervisor/hyperpod.go
```diff
@@ -123,7 +123,12 @@ func (hp *HyperPod) initPodNetwork(c *Container) error {
 			Ip:     info.Ip,
 		}
 
-		hp.vm.AddNic(info.Index, fmt.Sprintf("eth%d", idx), conf)
+		err = hp.vm.AddNic(info.Index, fmt.Sprintf("eth%d", idx), conf)
+		if err != nil {
+			glog.Error(err)
+			return err
+		}
+
```
laijs: if the patchset need to be update, this check can be moved back to the patch which add `hp.vm.AddNic(....)`

containerd/nslistener.go
```diff
@@ -83,6 +83,8 @@ func collectionInterfaceInfo() []supervisor.InterfaceInfo {
 			glog.Infof("get interface %v", info)
 			infos = append(infos, info)
 		}
+		// set link down, tap device take over it
+		netlink.LinkSetDown(link)
```
laijs: does  the docker daemon try to retrieve any information in some cases when the container is running?

PR #228, opened 2016-05-19
mv DecodedMessage in hypervisor/init_comm.go to hyperstart/api/json/types.go

YaoZengzeng:


hypervisor/init_comm.go
```diff
@@ -55,7 +49,7 @@ func waitConsoleOutput(ctx *VmContext) {
 	}
 }
 
-func NewVmMessage(m *DecodedMessage) []byte {
+func NewVmMessage(m *hyperstartapi.DecodedMessage) []byte {
```
laijs: It seems better if NewVmMessage() is also moved with a proper name. any suggestion?

hypervisor/init_comm.go
```diff
@@ -64,7 +58,7 @@ func NewVmMessage(m *DecodedMessage) []byte {
 	return msg
 }
 
-func ReadVmMessage(conn *net.UnixConn) (*DecodedMessage, error) {
+func ReadVmMessage(conn *net.UnixConn) (*hyperstartapi.DecodedMessage, error) {
```
laijs: It seems better if a part of ReadVmMessage() is also moved with a proper name. any suggestion?

`func ReadDecodedMessage(r io.Reader) (*hyperstartapi.DecodedMessage, error)`

YaoZengzeng: If so, what about readTtyMessage and the ttyMessage send into tty.sock?

YaoZengzeng:
And why there are read methods and no write method ? I think it's confusing, we'd better encapsulate a write method, make read & write as API.Maybe its better to provide function rather than data struct as APIs, especially the function like read & write?

laijs:
read & write & the data struct all are API/ABI. please move them all if available.

laijs:
I merge this one, the read/write or binary encoding/decoding can be put in a subsequent PR.

PR #229, opened 2016-05-20
mv DecodedMessage related method to hyperstart/api/json/types.go

YaoZengzeng:


laijs:
close it, it is not strictly needed. it can be reopened when needed.

PR #230, opened 2016-05-21
use GenericOperation to simplify vm operation

gao-feng:


laijs:
please don't close the channel!

https://groups.google.com/forum/#!msg/golang-nuts/pZwdYRGxCIk/qpbHxRRPJdUJ
```
Note that it is only necessary to close a channel if the receiver is
looking for a close.  Closing the channel is a control signal on the
channel indicating that no more data follows.
```

hypervisor/vm.go
```diff
@@ -421,57 +421,64 @@ func (vm *Vm) KillContainer(container string, signal syscall.Signal) error {
...
-				return
-			}
-			result <- fmt.Errorf("allocate device failed")
-			return
-		default:
-			result <- fmt.Errorf("get unexpected event %s", EventString(ev.Event()))
-			return
-		}
+	client := make(chan VmEvent, 1)
+	defer close(client)
```
laijs: don't close the channel in receiver side.

laijs:
an example
```go
package promise

// Go is a basic promise implementation: it wraps calls a function in a goroutine,
// and returns a channel which will later return the function's return value.
func Go(f func() error) chan error {
	ch := make(chan error, 1)
	go func() {
		ch <- f()
	}()
	return ch
}
```

hypervisor/vm.go
gao-feng: get, will update

gao-feng:
do we really need promise for SendGenericOperation?

laijs:
promise is an example that the channel is not closed after used.

gao-feng:
ok, get it

PR #232, opened 2016-05-23
Temp container

bergwolf:
Support creating temporary container and waiting for container finish executing.

Work with https://github.com/hyperhq/hyperstart/pull/103 and https://github.com/hyperhq/hyperd/pull/328


bergwolf:
travis fails because this needs https://github.com/hyperhq/hyperstart/pull/103 to support tempContainer field. otherwise it fails with `get unknown section tempContainer in container`

PR #243, opened 2016-05-30
log for qemu

YaoZengzeng:


laijs:
see #277 
-D
and retrieve the stdout/stderr output of cmd.Run() and put them in glog

PR #260, opened 2016-06-16
Omit error when it has empty msg

resouer:
When there's no service defined in Pod spec, `ListService` will receive a error with empty msg returned by `ReadFile`, it's obscure and hard to debug.

How to reproduce:
`ListService` for any Pod without service discovery defined.

### DIFF
--- a/hypervisor/vm.go
+++ b/hypervisor/vm.go
@@ -16,6 +16,10 @@ import (
 	"github.com/hyperhq/runv/hypervisor/types"
 )
 
+const (
+	vmNoFileReadError = "Error: "
+)
+
 type Vm struct {
 	Id     string
 	Pod    *PodStatus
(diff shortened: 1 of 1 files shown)


hypervisor/vm.go
```diff
@@ -349,6 +353,11 @@ func (vm *Vm) ReadFile(container, target string) ([]byte, error) {
 		ctx.vm <- &cmd
 	}, StateRunning)
 
+	// if we got empty error, redefine it, it's a container without service
+	if err != nil && err.Error() == vmNoFileReadError {
+		err = fmt.Errorf("No service-discovery configure file found, this Pod has no service defined!")
```
resouer: will refactor this into empty svc list instead of error

gao-feng:
servicediscovery in Hyperd should handle this error. it's better to return error to the caller of readfile.
BTW, Can you show the logs? As I know hyperd doesn't read file when the pod has no service defined.

resouer:
@gao-feng In my previous commit in #369, I accidentally changed `status` to `spec` in this line:
```
if pod.status.Type != "service-discovery" || len(pod.status.Containers) <= 1 {
	return nil, "", fmt.Errorf("Pod %s doesn't have services discovery", podId)
}
``` 

I think that's why I received empty error, will close this one.

Issue #210, opened 2016-05-11
bug report: stdin/stdout hangs when call runv-containerd


WeiZhang555:
When I try to call runv-containerd with docker, if I don't specify terminal(-t), it hangs.

Normal case with `-t`:
```
$ docker run -ti --net none busybox ls
bin   dev   etc   home  proc  root  sys   tmp   usr   var
$
```

it hangs if without `-t`:
```
$ docker run -i --net none busybox ls
bin
dev
etc
home
proc
root
sys
tmp
usr
var

<hangs...Ctrl-C, Ctrl-D don't help>
```

laijs:
runv-containerd had  already reported the exit event and closed all the stdio fifo. (I don't know what is missing).
the docker daemon still hold the stdin fifo with rw access. so the docker attach can't end.

PR #265, opened 2016-06-18
Bug fix: runv-contianerd hangs when no terminal

WeiZhang555:
Fixes https://github.com/hyperhq/runv/issues/210

When trying to interact between docker and runv-containerd, if docker
specify no terminal, client can't exit and everything hangs.

This is caused by blocked stderr pipe, if docker specify no terminal,
docker will open reader side of stderr pipe, but runv-containerd never
open write side of stderr side, which causes everything to hanging
state.

### DIFF
--- a/hypervisor/tty.go
+++ b/hypervisor/tty.go
@@ -22,6 +22,7 @@ type WindowSize struct {
 type TtyIO struct {
 	Stdin     io.ReadCloser
 	Stdout    io.WriteCloser
+	Stderr    io.WriteCloser
 	ClientTag string
 	Callback  chan *types.VmResponse
 	ExitCode  uint8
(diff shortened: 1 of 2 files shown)


WeiZhang555:
Reproduce steps:

```
1. start runv-containerd:
$ ./runv-containerd --debug ....

2. start docker daemon:
$ dockerd -D --containerd /run/runv-containerd/containerd.sock ......

3. try to start a container without tty
$ docker run -i --name test busybox sh
pwd
/
exit 11
<hanging...>

4. in another terminal, try to stop or ps container
$ docker stop test
<hanging...>
$ docker ps
<hanging...>
```

Reference code of docker: https://github.com/docker/docker/blob/master/libcontainerd/process_linux.go#L49-L53

If container has a TTY, then its stderr is only `emptyReader`, but without TTY, it opens stderr pipe file and tries to read data from pipe. As runv-containerd never hands stderr pipe, docker hangs.

That's why `docker run -ti xxx` works fine but `docker run -i xxx` won't.

WeiZhang555:
Already fixed, close this.

PR #267, opened 2016-06-23
Support for removing container

YaoZengzeng:


laijs:
INIT_REMOVECONTAINER was already added in #394, since the hyperstart had already defined it at the very beginning.
But runv is still lack of the functions to use INIT_REMOVECONTAINER. Does this RemoveContainer() still needed? Is INIT_KILLCONTAINER(kill with signal SIGKILL) a workable workaround?

PR #269, opened 2016-06-27
use UserVolume to describe container image

bergwolf:
So that it can describe complex rootfs volume like an rbd based one.

Signed-off-by: Peng Tao <bergwolf@gmail.com>

### DIFF
--- a/hypervisor/devicemap.go
+++ b/hypervisor/devicemap.go
@@ -170,16 +170,25 @@ func (ctx *VmContext) setContainerInfo(index int, container *hyperstartapi.Conta
 	container.Initialize = info.Initialize
 
 	if info.Fstype == "dir" {
-		container.Image = info.Image
+		container.Image = info.Image.Source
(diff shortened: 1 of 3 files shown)


feiskyer:
@bergwolf What's mean by 'describe complex rootfs volume'? Do you mean to save rootfs info into container.Image?

In my opinion, an image ID is always expected for container.Image.

bergwolf:
@feiskyer this is to eliminate the difference between private and public runv about ContainerInfo.Image. For rbd based rootfs, we need to pass rbd related information in setContainerInfo() to construct rbd based block device. Therefore ContainerInfo.Image is expanded to pod.UserVolume, as is done in private runv.

bergwolf:
@feiskyer for rbd based rootfs, we need to expand ContainerInfo.Image to include rbd related information.

gao-feng:
LGTM

PR #280, opened 2016-07-04
check attachment before Detach

laijs:
it might have been detached by ctx.ptys.Close()


gnawux:
LGTM

PR #281, opened 2016-07-05
Use "qemu -D logfile" to get log from qemu

YaoZengzeng:


hypervisor/qemu/qemu.go
```diff
@@ -26,6 +26,7 @@ type QemuContext struct {
 	wdt         chan string
 	qmpSockName string
 	qemuPidFile string
+	qemuLogFile string
```
laijs: it seams qemuLogFile unneeded

PR #285, opened 2016-07-07
fix possible Hang on qemuContext.Close and GetVm

gao-feng:


gao-feng:
travis failed due to container env is missed in exec, hyperd needs to implement this.

PR #286, opened 2016-07-11
Enable CNI

HuKeping:
The test script would be:
```
#! /bin/bash

set -x
set -e

# it is unlikely we'd use the notary
unset DOCKER_CONTENT_TRUST

# contaienrID of course.
CID=`docker run -d ubuntu:14.04 sleep 10000`

PID=`docker inspect -f "{{.State.Pid}}" $CID`

DEVICE_NAME="eth1"

# Use the bridge Ip, "docker0" for example, as the default gateway for container
Bridge=docker0
#BridgeIP=`ifconfig $Bridge | grep inet | grep -v inet6 | awk '{print $2}' |tr -d "addr:"`
BridgeIP="172.17.0.1"


mkdir -p /var/run/netns

# /var/run/netns/xxx is what `ip` wanted.
ln -s /proc/$PID/ns/net /var/run/netns/$PID

# set up veth pair A and B, attach A to bridge-docker0.
ip link add A type veth peer name B
brctl addif $Bridge A
ip link set A up

# add B into container's network namespace and hack it.
ip link set B netns $PID
ip netns exec $PID ip link set dev B name $DEVICE_NAME
ip netns exec $PID ip link set $DEVICE_NAME address 12:34:56:78:9a:bb
ip netns exec $PID ip addr add 172.17.42.101/16 dev $DEVICE_NAME
ip netns exec $PID ip link set $DEVICE_NAME up
#ip netns exec $PID ip route add default via $BridgeIP

# just in case we want do some thing
sleep 30

ip netns exec $PID ip link delete $DEVICE_NAME
sleep 30

# teardown
docker stop $CID
docker rm $CID

find -L /var/run/netns -type l -delete
ip link delete A
```

### DIFF
--- a/hypervisor/devicemap.go
+++ b/hypervisor/devicemap.go
@@ -489,6 +489,28 @@ func (ctx *VmContext) releaseNetwork() {
 	}
 }
 
+func (ctx *VmContext) releaseNetworkByLinkIndex(index int) {
+	var maps []pod.UserContainerPort
+
+	for _, c := range ctx.userSpec.Containers {
+		for _, m := range c.Ports {
+			maps = append(maps, m)
+		}
+	}
+
+	nic, ok := ctx.devices.networkMap[index]
+	if !ok {
+		glog.Error("trying to remove an un exist card:", nic)
+		return
+	}
+
+	if ctx.progress.deleting.networks[index] == false {
+		glog.V(1).Infof("remove network card %d: %s", index, nic.IpAddr)
+		go ctx.ReleaseInterface(index, nic.IpAddr, nic.Fd, maps)
+		maps = nil
+	}
+}
+
 func (ctx *VmContext) removeInterface() {
 	var maps []pod.UserContainerPort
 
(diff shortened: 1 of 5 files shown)


supervisor/hyperpod.go
```diff
@@ -153,15 +181,85 @@ func (hp *HyperPod) initPodNetwork(c *Container) error {
...
+				glog.Error(err)
+				continue
+			}
+
+			conf := pod.UserInterface{
+				Bridge: bridge,
+				Ip:     update.Addr.LinkAddress.String(),
+			}
+
+			err = hp.vm.AddNic(update.Addr.LinkIndex, "eth1", conf)
```
HuKeping: It is for test here, will change in the future.

hypervisor/devicemap.go
```diff
@@ -506,6 +524,19 @@ func (ctx *VmContext) removeInterface() {
 	}
 }
 
+func (ctx *VmContext) removeInterfaceByLinkIndex(index int) {
+	for idx, nic := range ctx.devices.networkMap {
+
+		glog.V(1).Infof("==============>idx:nic is %v:%v", idx, nic)
```
HuKeping: This is used to see what we have in the networkMap, it's a debug print, will remove in the future.

hypervisor/devicemap.go
```diff
@@ -489,6 +489,24 @@ func (ctx *VmContext) releaseNetwork() {
...
+func (ctx *VmContext) releaseNetworkByLinkIndex(index int) {
+	var maps []pod.UserContainerPort
+
+	for _, c := range ctx.userSpec.Containers {
+		for _, m := range c.Ports {
+			maps = append(maps, m)
+		}
+	}
+
+	for idx, nic := range ctx.devices.networkMap {
```
gao-feng: Don't need cycle here,
nic, ok := ctx.devices.networkMap[index]

hypervisor/devicemap.go
```diff
@@ -506,6 +524,19 @@ func (ctx *VmContext) removeInterface() {
 	}
 }
 
+func (ctx *VmContext) removeInterfaceByLinkIndex(index int) {
+	for idx, nic := range ctx.devices.networkMap {
```
gao-feng: same

hypervisor/devicemap.go
HuKeping: updated

HuKeping:
Updated.

gao-feng:
LGTM

PR #298, opened 2016-08-04
introduce struct QemuLogFile and watch it

YaoZengzeng:

### DIFF
--- a/hypervisor/qemu/qemu.go
+++ b/hypervisor/qemu/qemu.go
@@ -26,7 +26,7 @@ type QemuContext struct {
 	wdt         chan string
 	qmpSockName string
 	qemuPidFile string
-	qemuLogFile string
+	qemuLogFile *QemuLogFile
 	cpus        int
 	process     *os.Process
 }
@@ -51,9 +51,13 @@ func (qd *QemuDriver) InitContext(homeDir string) hypervisor.DriverContext {
 		os.Mkdir(QemuLogDir, 0755)
 	}
 
-	qemuLogFile := QemuLogDir + "/" + homeDir[strings.Index(homeDir, "vm-"):len(homeDir)-1] + ".log"
-	if _, err := os.Create(qemuLogFile); err != nil {
+	logFile := QemuLogDir + "/" + homeDir[strings.Index(homeDir, "vm-"):len(homeDir)-1] + ".log"
+	if _, err := os.Create(logFile); err != nil {
 		glog.Errorf("create qemu log file failed: %v", err)
+	}
+	qemuLogFile := &QemuLogFile{
+		Name:   logFile,
+		Offset: 0,
 	}
 
 	return &QemuContext{
@@ -74,6 +78,7 @@ func (qd *QemuDriver) LoadContext(persisted map[string]interface{}) (hypervisor.
 	}
 
 	var sock string
+	var log QemuLogFile
 	var proc *os.Process = nil
 	var err error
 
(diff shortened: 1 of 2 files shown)


hypervisor/qemu/qemu_process.go
```diff
@@ -13,6 +14,78 @@ import (
...
+	offset     int64
+	cancelChan chan bool
+}
+
+func (f *QemuLogFile) Read(p []byte) (n int, err error) {
+	reader, err := os.Open(f.name)
+	defer reader.Close()
+	if err != nil {
+		return 0, err
+	}
```
laijs: the defer and the if should be swapped.

and the `reader` is pushed to the sacked when the defer is invoked, but the reader can be changed after timeout, so you should wrapper the `reader.Close()` into a closure to delay the push the `reader`
```go
defer func() {
   reader.Close()
} ()
```

or just put the `reader.Close()` before every return.

hypervisor/qemu/qemu_process.go
```diff
@@ -13,6 +14,78 @@ import (
...
+	reader.Seek(f.offset, os.SEEK_SET)
+
+	for {
+		n, err = reader.Read(p)
+		f.offset += int64(n)
+
+		if err == io.EOF {
```
laijs: `if n == 0 && err == io.EOF {`

hypervisor/qemu/qemu.go
```diff
@@ -51,10 +51,15 @@ func (qd *QemuDriver) InitContext(homeDir string) hypervisor.DriverContext {
 		os.Mkdir(QemuLogDir, 0755)
 	}
 
-	qemuLogFile := QemuLogDir + "/" + homeDir[strings.Index(homeDir, "vm-"):len(homeDir)-1] + ".log"
-	if _, err := os.Create(qemuLogFile); err != nil {
+	logFile := QemuLogDir + "/" + homeDir[strings.Index(homeDir, "vm-"):len(homeDir)-1] + ".log"
+	if _, err := os.Create(logFile); err != nil {
```
laijs: the os.Create() will truncate it if it already exists, it is good for vm-name-recycle but bad for vm association.
it must be fixed.

laijs: sorry, it is not called from vm-association.
qemuLogFile need to be set in the LoadContext()

hypervisor/qemu/qemu_process.go
```diff
@@ -13,6 +14,80 @@ import (
...
+		reader, err := os.Open(f.name)
+		if err != nil {
+			return
+		}
+		f.offset, err = reader.Seek(0, os.SEEK_END)
+		if err != nil {
+			return
+		}
+		reader.Close()
+	}
```
laijs: the value of offset is saved and restored when associateQemu(), `skipAll bool` is unneeded, please remove it.

hypervisor/qemu/qemu_process.go
```diff
@@ -13,6 +14,68 @@ import (
 	"github.com/hyperhq/runv/hypervisor"
 )
 
+type QemuLogFile struct {
+	name   string
+	offset int64
```
laijs: the first letter of name and offset should be upper case for json marshal/unmarshal

hypervisor/qemu/qemu_process.go
```diff
@@ -13,6 +14,68 @@ import (
 	"github.com/hyperhq/runv/hypervisor"
 )
 
+type QemuLogFile struct {
+	Name   string `json:"name"`
+	Offset int64  `json:"offset"`
+	Eof    bool   `json:"eof"`
```
laijs: eof should be false when associateQemu()
so the eof shouldn't be saved.

PR #302, opened 2016-08-19
use `runv containerd --solo-namespaced` to replace `runv-namespaced`

laijs:


laijs:
test passed, merge it.

PR #312, opened 2016-08-29
add support for other platforms

pmorjan:
This patch adds initial support for s390x and ppc64le. The idea is to put all platform specific code into separated files for conditional compilation with the regular go build tool.
Currently only a few functions related to qemu execution need to be adjusted. That's great.

This pr is mainly to explore options. We are getting good results on s390x running regular Docker container. ppc64le is also on it's way.
Thanks.
peter

### DIFF
--- a/hypervisor/qemu/qemu.go
+++ b/hypervisor/qemu/qemu.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"os/exec"
 	"strconv"
 	"strings"
 
@@ -33,17 +32,6 @@ type QemuContext struct {
 
 func qemuContext(ctx *hypervisor.VmContext) *QemuContext {
 	return ctx.DCtx.(*QemuContext)
-}
-
-func InitDriver() *QemuDriver {
-	cmd, err := exec.LookPath("qemu-system-x86_64")
-	if err != nil {
-		return nil
-	}
-
-	return &QemuDriver{
-		executable: cmd,
-	}
 }
 
 func (qd *QemuDriver) Name() string {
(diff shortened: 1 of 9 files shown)


gnawux:
Thank you @pmorjan, great job.

hypervisor/qemu/qemu_ppc64le.go
```diff
@@ -0,0 +1,63 @@
...
+	var memParams, cpuParams string
+	if ctx.Boot.HotAddCpuMem {
+		memParams = fmt.Sprintf("size=%d,slots=1,maxmem=%dM", ctx.Boot.Memory, hypervisor.DefaultMaxMem) // TODO set maxmem to the total memory of the system
+		cpuParams = fmt.Sprintf("cpus=%d,maxcpus=%d", ctx.Boot.CPU, hypervisor.DefaultMaxCpus)           // TODO set it to the cpus of the system
+	} else {
+		memParams = strconv.Itoa(ctx.Boot.Memory)
+		cpuParams = strconv.Itoa(ctx.Boot.CPU)
+	}
+	memParams = "256"
+	cpuParams = "1"
```
laijs: it overwrite the results of the above code.

hypervisor/qemu/qemu_ppc64le.go
```diff
@@ -0,0 +1,63 @@
...
+			Initrd: hypervisor.DefaultInitrd,
+		}
+	}
+	boot := ctx.Boot
+	qc.cpus = boot.CPU
+
+	var memParams, cpuParams string
+	if ctx.Boot.HotAddCpuMem {
+		memParams = fmt.Sprintf("size=%d,slots=1,maxmem=%dM", ctx.Boot.Memory, hypervisor.DefaultMaxMem) // TODO set maxmem to the total memory of the system
+		cpuParams = fmt.Sprintf("cpus=%d,maxcpus=%d", ctx.Boot.CPU, hypervisor.DefaultMaxCpus)           // TODO set it to the cpus of the system
```
laijs: you can use boot.Memory instead of ctx.Boot.Memory,
boot.CPU instead of ctx.Boot.CPU.

hypervisor/qemu/qemu.go
```diff
@@ -35,17 +34,6 @@ func qemuContext(ctx *hypervisor.VmContext) *QemuContext {
 	return ctx.DCtx.(*QemuContext)
 }
 
-func InitDriver() *QemuDriver {
-	cmd, err := exec.LookPath("qemu-system-x86_64")
```
laijs: could you define multiple QEMU_SYSTEM_EXE in different arch-specific files
rather than define multiple InitDriver() in arch-specific files?

pmorjan: Yes this makes all sense. I'll have a look.

hypervisor/qemu/qmp_wrapper_s390x.go
```diff
@@ -0,0 +1,47 @@
...
+			"type": "tap", "id": device, "fd": "fd" + device,
+		},
+	}
+	commands[2] = &QmpCommand{
+		Execute: "device_add",
+		Arguments: map[string]interface{}{
+			"driver": "virtio-net-ccw",
+			"netdev": device,
+			"mac":    mac,
+			"id":     device,
```
laijs: I think there must be a "bus" for the network.
and "addr"(busaddr) is essential to identify the network device.

pmorjan: "virtio-net-ccw" is a mainframe channel device. Doesn't require bus name/address like PCI devices.

hypervisor/qemu/qemu.go
pmorjan: Two options: 
Three new files for arch specific constants as in this  [branch testing](https://github.com/hyperhq/runv/compare/master...pmorjan:testing?diff=split&name=testing) or 

new const declaration in the existing arch specific qemu_<arch>.go files as in this [branch testing2](https://github.com/hyperhq/runv/compare/master...pmorjan:testing2)

Not sure which one to prefer.

laijs: I prefer the second one.
thanks and it is very nice for you to provide both ways.

PR #316, opened 2016-09-01
POD support: Use ns-listerner pid as containers' pid

WeiZhang555:
For the sake of K8S integration, we must have a mechanism to create a
pod and allow to integrate with CNI network plugins.

As CNI plugins are operating on nslistener namespaces, we should set pid
of nslistener as container's init pid, so that when we insert a NIC into
the "fake" namespace, the VM can get notifications.

### DIFF
--- a/supervisor/hyperpod.go
+++ b/supervisor/hyperpod.go
@@ -390,7 +390,7 @@ func (hp *HyperPod) createContainer(container, bundlePath, stdin, stdout, stderr
 		Stdout: stdout,
 		Stderr: stderr,
 		Spec:   &spec.Process,
-		ProcId: -1,
+		ProcId: c.ownerPod.getNsPid(),
 
 		inerId:    inerProcessId,
 		ownerCont: c,
--- a/supervisor/supervisor.go
+++ b/supervisor/supervisor.go
@@ -3,7 +3,8 @@ package supervisor
 import (
 	"fmt"
 	"os"
-	"strings"
+	"regexp"
+	"strconv"
 	"sync"
 	"time"
 
@@ -141,23 +142,30 @@ func (sv *Supervisor) getHyperPod(container string, spec *specs.Spec) (hp *Hyper
 		return nil, fmt.Errorf("The container %s is already existing", container)
 	}
 	for _, ns := range spec.Linux.Namespaces {
-		if ns.Path != "" {
-			if strings.Contains(ns.Path, "/") {
-				return nil, fmt.Errorf("Runv doesn't support path to namespace file, it supports containers name as shared namespaces only")
-			}
+		if len(ns.Path) > 0 {
 			if ns.Type == "mount" {
 				// TODO support it!
 				return nil, fmt.Errorf("Runv doesn't support shared mount namespace currently")
 			}
-			shared := ns.Path
-			cnt, ok := sv.Containers[shared]
-			if !ok {
-				return nil, fmt.Errorf("The container %s is not existing to share with", shared)
+
+			pidexp := regexp.MustCompile(`/proc/(\d+)/ns/*`)
+			matches := pidexp.FindStringSubmatch(ns.Path)
+			if len(matches) != 2 {
+				return nil, fmt.Errorf("Can't find shared container with network ns path %s", ns.Path)
+			}
+			pid, _ := strconv.Atoi(matches[1])
+
+			for _, c := range sv.Containers {
+				if c.ownerPod != nil && pid == c.ownerPod.getNsPid() {
+					if hp != nil && hp != c.ownerPod {
+						return nil, fmt.Errorf("Conflict share")
+					}
+					hp = c.ownerPod
+					break
+				}
 			}
 			if hp == nil {
-				hp = cnt.ownerPod
-			} else if cnt.ownerPod != hp {
-				return nil, fmt.Errorf("conflict share")
+				return nil, fmt.Errorf("Can't find shared container with network ns path %s", ns.Path)
 			}
 		}
 	}


WeiZhang555:
From now on, runv can support "POD"again. To compose a POD, you need to run something like:
```
# docker run -tid --name test1 busybox sh
# docker run -tid --net container:test1 busybox sh   // composing a pod
```
Note that only `--net` can making a POD, `--ipc container:xxx`, `--pid container:xxx` will be ignored.

Using `docker top` to check two containers in one pod, you will get result like this:
```
$ docker top 2db792e747a3
UID                 PID                 PPID                C                   STIME               TTY                 TIME                CMD
root                23612               23416               0                   14:13               pts/24              00:00:00            containerd-nslistener
$ docker top 448783d97d38
UID                 PID                 PPID                C                   STIME               TTY                 TIME                CMD
root                23612               23416               0                   14:13               pts/24              00:00:00            containerd-nslistener
```

WeiZhang555:
The test case failed because log is too long, I think you should disable debug flag for hyperd test cases.

supervisor/supervisor.go
```diff
@@ -141,24 +141,24 @@ func (sv *Supervisor) getHyperPod(container string, spec *specs.Spec) (hp *Hyper
 		return nil, fmt.Errorf("The container %s is already existing", container)
 	}
 	for _, ns := range spec.Linux.Namespaces {
-		if ns.Path != "" {
-			if strings.Contains(ns.Path, "/") {
-				return nil, fmt.Errorf("Runv doesn't support path to namespace file, it supports containers name as shared namespaces only")
+		if ns.Type == "network" && len(ns.Path) > 0 {
```
laijs: don't limit it on network namespace

supervisor/supervisor.go
```diff
@@ -141,24 +141,24 @@ func (sv *Supervisor) getHyperPod(container string, spec *specs.Spec) (hp *Hyper
...
+					break
+				}
 			}
 			if hp == nil {
-				hp = cnt.ownerPod
-			} else if cnt.ownerPod != hp {
-				return nil, fmt.Errorf("conflict share")
+				return nil, fmt.Errorf("Can't find shared container with network ns path %s", ns.Path)
 			}
+			continue
```
laijs: why this continue is needed?

supervisor/supervisor.go
```diff
@@ -141,24 +141,24 @@ func (sv *Supervisor) getHyperPod(container string, spec *specs.Spec) (hp *Hyper
...
 			}
-			if ns.Type == "mount" {
-				// TODO support it!
-				return nil, fmt.Errorf("Runv doesn't support shared mount namespace currently")
-			}
-			shared := ns.Path
-			cnt, ok := sv.Containers[shared]
-			if !ok {
-				return nil, fmt.Errorf("The container %s is not existing to share with", shared)
+			pid := matches[1]
```
laijs: convert it to int

supervisor/supervisor.go
WeiZhang555: I think we have to. 
Imagine this "--net container:a --pid container:b --ipc container:c", we can't make a logical decision on which POD to join.

POD is concept from K8S and K8S will always join namespaces of same container, I think limiting it to network namespace will work fine.

supervisor/supervisor.go
WeiZhang555: Right, I'll remove it.

supervisor/supervisor.go
WeiZhang555: `pid` is only used for comparation, e.g. `pid == fmt.Sprintf("%d", c.ownerPod.getNsPid())`

So I think either converting `pid` to int or converting `c.ownerPod.getNsPid()` to string could work fine, they are basically the same :smile:

supervisor/supervisor.go
laijs: report an error when conflict

```go
 			if hp == nil {
 				hp = cnt.ownerPod
 			} else if cnt.ownerPod != hp {
 				return nil, fmt.Errorf("conflict share")
```

supervisor/supervisor.go
```diff
@@ -141,23 +141,22 @@ func (sv *Supervisor) getHyperPod(container string, spec *specs.Spec) (hp *Hyper
...
+
+			for _, c := range sv.Containers {
+				if c.ownerPod != nil && c.ownerPod.getNsPid() > 0 && pid == fmt.Sprintf("%d", c.ownerPod.getNsPid()) {
```
laijs: it need to be shorten

gao-feng: no need to check c.ownerPod.getNsPid() > 0

WeiZhang555: right :+1:

supervisor/supervisor.go
```diff
@@ -141,23 +141,22 @@ func (sv *Supervisor) getHyperPod(container string, spec *specs.Spec) (hp *Hyper
 		return nil, fmt.Errorf("The container %s is already existing", container)
 	}
 	for _, ns := range spec.Linux.Namespaces {
-		if ns.Path != "" {
-			if strings.Contains(ns.Path, "/") {
-				return nil, fmt.Errorf("Runv doesn't support path to namespace file, it supports containers name as shared namespaces only")
+		if ns.Type == "network" && len(ns.Path) > 0 {
+			pidexp := regexp.MustCompile(`/proc/(\d+)/ns/net`)
```
gao-feng: Please check if this path is always ns files under the /proc, if not, please compare the inode of two files.

WeiZhang555: At least in docker scenarios, it's always "/proc/xxx"

WeiZhang555:
Updated.

supervisor/supervisor.go
```diff
@@ -141,23 +142,25 @@ func (sv *Supervisor) getHyperPod(container string, spec *specs.Spec) (hp *Hyper
...
+		if len(ns.Path) > 0 {
+			pidexp := regexp.MustCompile(`/proc/(\d+)/ns/*`)
+			matches := pidexp.FindStringSubmatch(ns.Path)
+			if len(matches) != 2 {
+				return nil, fmt.Errorf("Can't find shared container with network ns path %s", ns.Path)
 			}
-			if ns.Type == "mount" {
-				// TODO support it!
-				return nil, fmt.Errorf("Runv doesn't support shared mount namespace currently")
-			}
```
laijs: don't remove this check

supervisor/hyperpod.go
```diff
@@ -408,6 +409,20 @@ func (hp *HyperPod) createContainer(container, bundlePath, stdin, stdout, stderr
...
+	alarm := time.After(5 * time.Second)
+	for {
+		if pid := c.ownerPod.getNsPid(); pid > 0 {
+			p.ProcId = pid
+			break
+		}
+
+		select {
+		case <-alarm:
+			glog.V(3).Infof("Timeout while waiting for container's pid")
```
laijs: do we need to break the for loop?

supervisor/hyperpod.go
```diff
@@ -408,6 +409,20 @@ func (hp *HyperPod) createContainer(container, bundlePath, stdin, stdout, stderr
...
+		default:
+			time.Sleep(100 * time.Millisecond)
+		}
+	}
```
laijs: what's the purpose of "waiting for container's pid",
the proxy pid was created just after the hyperpod was created.
we don't need to wait it.

laijs: correct me if I'm wrong. @gao-feng

supervisor/supervisor.go
WeiZhang555: I have to. When we were trying to integrate runv-containerd with K8S, it seems K8S will always fail on this. K8S specified a mount namespace path.

I didn't look into the detail of K8S, but it seems keep this check will make K8S failed of makeing POD.

supervisor/hyperpod.go
WeiZhang555: Right

supervisor/hyperpod.go
WeiZhang555: The line "c.run(p)" above this piece of code is put into a goroutine, the creating process is also put into the goroutine, we have to wait because nothing is done when this "waiting" is executed.

laijs: `sv.CreateContainer() -> sv.getHyperPod() -> createHyperPod() -> hp.startNsListener() -> cmd.Start()` all are directly calls, after sv.CreateContainer() returns, the proxy/pid was created.
`hp.createContainer()` is after `sv.getHyperPod()` so the code in `hp.createContainer()` can sees the proxy/pid.

supervisor/supervisor.go
laijs: but runv doesn't fully take care of mount share yet, it might include various bugs.

WeiZhang555: If I added this check back, K8S will fail to interact with runv. 
I'm sure I must remove this in our internal version. But if you insist, I can add it back only in this upstream version

supervisor/hyperpod.go
WeiZhang555: Oh, it seems you are right :+1:

supervisor/supervisor.go
WeiZhang555: I'll add this back and maintain a different internal version.

laijs: do you check the possible bugs when containers share mount namespace ?
if you fix it internal, please update it in this repo.
if you have enough confidence that it works well with share mount namespace(without this check), could you fire another pr to remove this check?

WeiZhang555: I'm not sure what problem it will bring if I remove this check, we have an internal K8S, it works fine with this for some time.

laijs: how about test it with pure runv commands?

WeiZhang555: pure runv commands will report 
`Runv doesn't support path to namespace file, it supports containers name as shared namespaces only`

It's reported by https://github.com/hyperhq/runv/blob/master/start.go#L93, so currently only runv-containerd can support namespace path, and it won't bring impact on pure runv command.

PR #323, opened 2016-09-08
integration test for list -f json

YaoZengzeng:

### DIFF
--- a/integration-test/list_test.go
+++ b/integration-test/list_test.go
@@ -1,11 +1,24 @@
 package main
 
 import (
+	"encoding/json"
+	"strings"
 	"time"
 
 	"github.com/docker/docker/pkg/integration/checker"
 	"github.com/go-check/check"
+	"github.com/opencontainers/runtime-spec/specs-go"
 )
+
+type containerState struct {
+	specs.State
+	// Bundle is the path on the filesystem to the bundle
+	Bundle string `json:"bundle"`
+	// Status is the current status of the container, running, paused, ...
+	Status string `json:"status"`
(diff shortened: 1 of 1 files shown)


integration-test/list_test.go
```diff
@@ -1,12 +1,27 @@
...
+	"encoding/json"
+	"strings"
 	"time"
 
 	"github.com/docker/docker/pkg/integration/checker"
 	"github.com/go-check/check"
 )
 
+type cList struct {
+	// ID is the container ID
```
gao-feng: Is this structure from containerState of runc?
I suggest we use State structure from oci. Create new structure combinded with the different fields and State.
Such as
```
type containerState struct {
    specs.State
    Bundle string `json:"bundle"` 
}
```

integration-test/list_test.go
```diff
@@ -36,3 +51,49 @@ func (s *RunVSuite) TestListSleep(c *check.C) {
 	c.Assert(out, checker.Contains, ctrName)
 	<-exitChan
 }
+
+func (s *RunVSuite) TestListSleepJson(c *check.C) {
+	//TODO: enable this after fixing!!!
+	//c.Skip("enable this after fixing")
```
gao-feng: what's this meaning?

YaoZengzeng: Hmmm... I don't know as well.
I just find every test case has it at the beginning, so I write it there.

YaoZengzeng:
Ready for review

gao-feng:
lgtm

PR #335, opened 2016-09-19
report error when the workdir in spec is empty.

YaoZengzeng:
report error when the workdir in spec is empty.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -153,6 +153,9 @@ func getProcess(context *cli.Context, bundle string) (*specs.Process, error) {
 	}
 	p := spec.Process
 	p.Args = context.Args()[1:]
+	if p.Cwd == "" {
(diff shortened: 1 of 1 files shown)


laijs:
in oci, it is required field, we should also make it required.

exec.go
```diff
@@ -154,6 +154,10 @@ func getProcess(context *cli.Context, bundle string) (*specs.Process, error) {
 	}
 	p := spec.Process
 	p.Args = context.Args()[1:]
+	// if cwd in config.json is "", set it as "/"
+	if p.Cwd == "" {
+		p.Cwd = "/"
+	}
```
laijs: please report error

laijs:
```go
        if p.Cwd == "" {
                return fmt.Errorf("Cwd property must not be empty")
        }
```

PR #338, opened 2016-09-19
replace busybox.tar binary by Makefile

pmorjan:
This pr adds a Makefile to replace  the binary [integration-test/test_data/busbox.tar](../blob/master/integration-test/test_data/busybox.tar)

The old busybox.tar also includes '/usr/bin/getconf' which is not part of busybox. I'm not sure if this is still needed. The new Makefile builds this a well. Could be removed if it's no longer needed.


laijs:
would you try to remove '/usr/bin/getconf' and test?

pmorjan:
looks good to me

PR #340, opened 2016-09-20
Do not handle stop pod event in state machine

gao-feng:
Stop pod event is abandoned


laijs:
rename INIT_STOPPOD to INIT_STOPPOD_DEPRECATED

PR #348, opened 2016-09-22
[minor] Correct some misspell in comments or const string

gnawux:


gnawux:
I confirm this patch did not change any code, only comments and const strings for log or print

PR #350, opened 2016-09-22
set current directory as default bundle value

YaoZengzeng:

### DIFF
--- a/start.go
+++ b/start.go
@@ -27,6 +27,14 @@ func firstExistingFile(candidates []string) string {
 	return ""
 }
 
+func getDefaultBundlePath() string {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return ""
+	}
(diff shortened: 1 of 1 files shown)


start.go
```diff
@@ -46,7 +46,7 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name:  "bundle, b",
-			Usage: "path to the root of the bundle directory",
+			Usage: "path to the root of the bundle directory, defaults to the current directory",
```
laijs: directly set the value here.
```go
                     Value: getDefaultBundlePath(),
```

```go
func getDefaultBundlePath() string {
       cwd, err := os.Getwd()
       if err != nil {
               return ""
       }
       return cwd
}
```

PR #362, opened 2016-10-14
launch containerd through "/proc/self/exe containerd"

YaoZengzeng:
In cri-o, it use exec.Command(...) to launch runtime, it dosen't inherit PATH from parent process.
For runv, it can't launch containerd because  can't find  runv in PATH.
So use "/proc/self/exe" rather than "runv" to launch containerd.


laijs:
L164: path, err := osext.Executable()
L206: cmd.Path = path

Doesn't it work?

PR #364, opened 2016-10-19
Tiny fix to integrate with cri-o

Crazykev:
1. fix can't find runv when it is not in $PATH
2. do not support devpts mount type yet, skip it.

cc @laijs

### DIFF
--- a/start.go
+++ b/start.go
@@ -202,8 +202,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 				"--state-dir", root,
 				"--listen", filepath.Join(namespace, "namespaced.sock"),
 			)
-			cmd = exec.Command("runv", args...)
-			cmd.Path = path
+			cmd = &exec.Cmd{
+				Path: path,
+				Args: append([]string{"runv"}, args...),
+			}
 			cmd.Dir = "/"
 			cmd.SysProcAttr = &syscall.SysProcAttr{
 				Setsid: true,
--- a/supervisor/container.go
+++ b/supervisor/container.go
@@ -312,7 +312,7 @@ func mountToRootfs(m *specs.Mount, rootfs, mountLabel string) error {
 	}
 
 	switch m.Type {
-	case "proc", "sysfs", "mqueue", "tmpfs", "cgroup":
+	case "proc", "sysfs", "mqueue", "tmpfs", "cgroup", "devpts":
 		glog.V(3).Infof("Skip mount point %q of type %s", m.Destination, m.Type)
 		return nil
 	case "bind":


start.go
```diff
@@ -202,8 +202,7 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 				"--state-dir", root,
 				"--listen", filepath.Join(namespace, "namespaced.sock"),
 			)
-			cmd = exec.Command("runv", args...)
-			cmd.Path = path
+			cmd = exec.Command(path, args...)
```
gao-feng: https://golang.org/src/os/exec/exec.go?s=4289:4334#L136
https://golang.org/src/os/exec/exec.go?s=4289:4334#L313

laijs: good.  @YaoZengzeng wdyt?

YaoZengzeng: @laijs @Crazykev  Acutally there is no need to do this.
Because of line164 "path, err := osext.Executable()",  the value of path is "/proc/self/exec", so it doesn't matter with runv is in PATH.
Although through my test it is OK if we don't inherit ENV from oci-d, shoudn't we cant find qemu?  it just gets abs through "cmd, err := exec.LookPath(QEMU_SYSTEM_EXE)" ?

YaoZengzeng: @laijs  Actually from the source code of golang os/exec package, we could find that  if cmd.Env is nil, system will transfer os.Environ to child by default.
[https://golang.org/src/os/exec/exec.go?s=4289:4334#L119](url)
[https://golang.org/src/os/exec/exec.go?s=4289:4334#L119](url)

Crazykev: @laijs yeah, we've already gurranteed the path is correct before, I'm OK with both, prefer this one.
```
cmd := &exec.Cmd{
            Path: path,
            Args: append([]string{"runv"}, args...),
        }
```
@YaoZengzeng 
> `cmd, err := exec.LookPath(QEMU_SYSTEM_EXE)` 

Also need $PATH to find the abs path if `QEMU_SYSTEM_EXE` is not. 
FYI: https://golang.org/src/os/exec/lp_unix.go

YaoZengzeng: @Crazykev  Yeah, as I metioned before, now in ocid they don't configure cmd.Env, we inherit all environments from ocid include PATH.
There will not be any problem for us to find qemu or any thing else in PATH, so everything is fine:)

Crazykev: @YaoZengzeng this one? https://github.com/kubernetes-incubator/cri-o/blob/master/oci/oci.go#L116

Crazykev:
@laijs Updated.

Crazykev:
Seems integration test or runv itself is not very stable

PR #365, opened 2016-10-20
Fix setWindowSize not working with docker

wrouesnel:
Docker sets "execId" to containerId + "-init" when working from the command line
with `docker run`. This led to `runv` being unable to find the referred
TTY context when setWindowSize messages were received.

This patch adds a check for the "-init" suffix and causes executors matching
this form to drop through to looking for the containerId only, which allows
setWindowSize to function correctly.

### DIFF
--- a/supervisor/process.go
+++ b/supervisor/process.go
@@ -64,6 +64,10 @@ func (p *Process) setupIO() error {
 }
 
 func (p *Process) ttyResize(container string, width, height int) error {
+	// If working on the primary process, do not pass execId (it won't be recognized)
+	if p.inerId == fmt.Sprintf("%s-init", container) {
+		return p.ownerCont.ownerPod.vm.Tty(container, "", height, width)
+	}
 	return p.ownerCont.ownerPod.vm.Tty(container, p.inerId, height, width)
 }
 


hypervisor/vm_states.go
```diff
@@ -153,7 +154,8 @@ func (ctx *VmContext) updateInterface(index int, result chan<- error) {
 
 func (ctx *VmContext) setWindowSize(containerId, execId string, size *WindowSize) {
 	var session uint64
-	if execId != "" {
+
+	if execId != "" && !strings.HasSuffix(execId, "-init") {
```
gao-feng: from your change, seems like containerId is not null? can you use `execId != containerId+"-init"` directly here?

wrouesnel: Yes that would also be the same. I'll update it.

wrouesnel:
Updated and tested working.

gao-feng:
@wrouesnel 

Sorry for confusing you, thought this a little more, VmContext.setWindowSize is the core part of runv, it has no relationship with docker, it's not quite proper place to handle the special case of docker.

Can you do this change in [Process.ttyResize](https://github.com/hyperhq/runv/blob/master/supervisor/process.go#L66)? if p.inerId equals to containerId+"-init", then pass "" as execId to vm.Tty.  supervisor is the interface that runv communicates
with docker.

Sorry again.

wrouesnel:
I see what you mean. I've updated the patch to fix it in the right location now.

gao-feng:
LGTM, thanks!

PR #368, opened 2016-10-21
fix readonly rootfs failed when creating necessary dirs

Crazykev:
Mainly fix readonly rootfs failed when creating necessary dirs

1. move mounting readonly rootfs to bottom
2. pre-create necessary dirs used in hyperstart

cc @laijs 


Crazykev:
@laijs updated.

laijs:
LGTM

Issue #373, opened 2016-10-24
runv kill dosen't clean up properly if have process other than init in container


YaoZengzeng:
First I start a container and exec in it

root@monster-Z:/home/monster/TEST/runv/container# runv start test
/ # ps
PID   USER     TIME   COMMAND
    1 root       0:00 /init
    3 root       0:00 sh
    4 root       0:00 ping 127.0.0.1
    5 root       0:00 ps

Then after I kill it , we can find it still alive through `runv list`

root@monster-Z:/home/monster/TEST/runv/container# runv list
ID          PID         STATUS      BUNDLE                              CREATED
test        6719        running     /home/monster/TEST/runv/container   2016-10-24T12:31:05.333658596+08:00
root@monster-Z:/home/monster/TEST/runv/container# runv kill test KILL
root@monster-Z:/home/monster/TEST/runv/container# runv list
ID          PID         STATUS      BUNDLE                              CREATED
test        6719        running     /home/monster/TEST/runv/container   2016-10-24T12:31:05.333658596+08:00

YaoZengzeng:
It seems when runv is connected with docker-containerd, there is no such problem.

laijs:
I can't reproduce the problem as this issues said.

Other processes will be killed in hyperstart(https://github.com/hyperhq/hyperstart/pull/96), the runv will receive the finished events, and cleanup the processes.

gnawux:
close this issue since it does not appear in current head.

PR #374, opened 2016-10-24
if the container's init process exit, clean up all other processes

YaoZengzeng:
fix issue #373 


laijs:
what if the container's init process exit by itself?
have you try docker of containerd in this case? we following them.

wrouesnel:
If the docker container's init exits, docker's behavior is to kill the rest of the container's processes.

I've been running with this patch for a while and it seems to work okay.

laijs:
other processes will be killed in hyperstart(https://github.com/hyperhq/hyperstart/pull/96), the runv will receive the finished event, and cleanup the processes. Is this pr really needed?

laijs:
the hyperstart does it

PR #375, opened 2016-10-24
implement `runv delete`

YaoZengzeng:
just a simple wrapper of runv kill container-id KILL

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -1,0 +1,56 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/codegangsta/cli"
+	"github.com/docker/containerd/api/grpc/types"
+	"github.com/hyperhq/runv/lib/linuxsignal"
+	netcontext "golang.org/x/net/context"
+)
+
+var deleteCommand = cli.Command{
+	Name:  "delete",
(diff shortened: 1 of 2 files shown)


Crazykev:
Will it fail if exec `runv delete` after `runv kill` ?

YaoZengzeng:
@Crazykev In fact, we can't distinguish a container being killed before and a container that never exist.
(1)、Now runv is stateless actually, the container is running or none.After kill, it just exited.
(2)、There in no api of deleting in containerd.So we can just do like this to work around.
Any good suggestions?

Crazykev:
AFAK, for a long term, we need to follow https://github.com/opencontainers/runtime-spec/pull/513 when it land. And in runv, maybe it's more easy to figure out what is the proper way to define `delete` after we seprerate `create` and `start`.  
And if this is only a workaround for ocid integration for now, that we need a `delete`. At least we need to gurantee `delete` after `kill` will not fail, that is why I would ask that.

gao-feng:
deleting container is  needed, please consider to implement it in containerd.

laijs:
@YaoZengzeng  @Crazykev runv cli is being improved continuously and had just been refactored at #537 .

Moreover, `runv delete` was also implemented in #537. Feel free to test ti and give any possible feedback.

PR #378, opened 2016-10-25
Improve supervisor eventlog locking and API compliance.

wrouesnel:
Add finer-grained locking to the supervisor eventlog based on the upstream docker containerd code.

Adds full compliance for the Event() API by supporting queries for specific container IDs and stored events only, as required by newer Docker daemons in order to recover from hard shutdowns without deadlocking the docker daemon.

This fixes the problem of deadlocking docker when running with runv and doing a hard shutdown, as runv was not responding with the events docker actually wanted.

The code is mostly adapted from https://github.com/docker/containerd/blob/master/supervisor/supervisor.go.

### DIFF
--- a/containerd/api/grpc/server/server.go
+++ b/containerd/api/grpc/server/server.go
@@ -154,7 +154,7 @@ func (s *apiServer) Events(r *types.EventsRequest, stream types.API_EventsServer
 		}
 		t = from
 	}
-	events := s.sv.Events.Events(t)
+	events := s.sv.Events.Events(t, r.StoredOnly, r.Id)
 	defer s.sv.Events.Unsubscribe(events)
 	for e := range events {
 		tsp, err := ptypes.TimestampProto(e.Timestamp)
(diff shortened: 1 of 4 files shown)


wrouesnel:
Hm, this actually looks like the issue in #379 is being made visible to tests now...

supervisor/container.go
```diff
@@ -30,9 +30,16 @@ type Container struct {
 	ownerPod *HyperPod
 }
 
-func (c *Container) run(p *Process) {
+func (c *Container) run(p *Process) error {
```
gao-feng: Changes in this file duplicate with https://github.com/hyperhq/runv/pull/379

wrouesnel:
This patch is built on the prior PR at this point, and not quite ready since it doesn't pass existing tests yet. I'm still playing around with the event log code to figure out how it should best work.

laijs:
@wrouesnel runv cli is being improved continuously and had just refactored #537 . As a result, it seams that the changes in this pr is outdated.

PR #379, opened 2016-10-26
Handle exec being called before a container is initialized correctly.

### LINKED ISSUE #279: runv-containerd hangs
Start runv-containerd and connect it with docker daemon, then execute: 

```
# cid=$(docker run -tid busybox sh) && docker exec -ti $cid echo hello 
```

docker daemon hangs. 

After some investigation, I found that at the time when it ran "docker exec", the container is actually not started/ready yet. 

I can throw an error out but that doesn't solve it. Could you fix it?

wrouesnel:
Running a command like the following:

```
cid=$(./docker run -tid busybox sh) && \
    echo $cid && ./docker exec -ti $cid echo hello
```

would block indefinitely because we would return success to docker, but fail because the exec is attempted before the container is started up.

This leads to the docker daemon blocking indefinitely waiting for an exec command it believes was successful via containerd.

This patch solves the problem by blocking the containerd responses until the VM startup has returned and moved to process wait (or else returning an error if it fails at this stage). This resolves deadlocks into docker and allows the above command to succeed.

Closes #279

### DIFF
--- a/supervisor/container.go
+++ b/supervisor/container.go
@@ -29,7 +29,9 @@ type Container struct {
 	ownerPod *HyperPod
 }
 
-func (c *Container) run(p *Process) {
+func (c *Container) run(p *Process) error {
+	resultCh := make(chan error)
+
 	go func() {
 		err := c.create(p)
 		if err != nil {
@@ -51,6 +53,9 @@ func (c *Container) run(p *Process) {
 		}
 		c.ownerPod.sv.Events.notifySubscribers(e)
 
+		resultCh <- err
+		close(resultCh)
+
 		exit, err := c.wait(p, res)
 		e = Event{
 			ID:        c.Id,
@@ -65,6 +70,9 @@ func (c *Container) run(p *Process) {
 		}
 		c.ownerPod.sv.Events.notifySubscribers(e)
 	}()
+
+	err := <-resultCh
+	return err
 }
 
 func (c *Container) create(p *Process) error {
--- a/supervisor/hyperpod.go
+++ b/supervisor/hyperpod.go
@@ -416,7 +416,9 @@ func (hp *HyperPod) createContainer(container, bundlePath, stdin, stdout, stderr
 	c.ownerPod.Containers[container] = c
 
 	glog.Infof("createContainer() calls c.run(p)")
-	c.run(p)
+	if err := c.run(p); err != nil {
+		return nil, err
+	}
 	return c, nil
 }
 


supervisor/container.go
```diff
@@ -226,30 +236,41 @@ func (c *Container) addProcess(processId, stdin, stdout, stderr string, spec *sp
...
-		PID:       processId,
-	}
-	c.ownerPod.sv.Events.notifySubscribers(e)
+	// Wait for the VM to acknowledge the process has attached to avoid a
+	// libcontainer deadlock.
+	resultCh := make(chan error)
 
 	go func() {
 		c.ownerPod.podStatus.AddExec(c.Id, processId, "", spec.Terminal)
 		err := c.ownerPod.vm.AddProcess(c.Id, processId, spec.Terminal, spec.Args, spec.Env, spec.Cwd, p.stdio)
```
gao-feng: vm.AddProcess will block until process is finished, this change will block AddProcess. And seems like we double call the WaitForFinish, this is unexpected. 

IMO, EventProcessStart & EventExit seems in the right order & place. even vm.AddProcess failed, we will send out EventExit.

supervisor/container.go
```diff
@@ -30,9 +30,16 @@ type Container struct {
 	ownerPod *HyperPod
 }
 
-func (c *Container) run(p *Process) {
+func (c *Container) run(p *Process) error {
+	// Receives the early result of the attempt to start the container,
+	// which ensures that we do not return before EventContainerStart
+	// is emitted.
```
gao-feng: Yes, you are right, this part looks good to me, one tips, we can move c.start out of goroutine and leave c.wait inside goroutine to get rid of channel. 

bad news is vm.NewContainer in c.start doesn't guarantee container is started in vm, it just passes out the `new container` message, we still need another patch to fix this.

wrouesnel: I've noticed that - this fix is still racey in that regard, but it seems to mostly work. In general the problem seems to be there's a missing "container started" message that hyperstart needs to be returning.

supervisor/container.go
wrouesnel: This was motivated by it being unclear what the expectation higher up containerd is if AddProcess fails. If we fail to ever start the process, then an attempt at recovering containers later isn't going to find it since it was probably a runv/hyperstart error. So better not to notify it was added until we're sure it was.

supervisor/container.go
gao-feng: we has the "container started" message in hyperstart as "ack to new container" message, but runv doesn't wait for it.

supervisor/container.go
gao-feng: unfortunately vm.AddProcess exits after process is exited in current codes. the EventProcessStart in this pr is sent  out too late

wrouesnel: I'll see if I can concoct a test to catch this situation so I can avoid it (my point testing was based on the noted ticket, which of course exits quite quickly).

supervisor/container.go
wrouesnel: I'll revisit this code in that case, as this might require a more elaborate change to be a proper fix (and I'd prefer to fix it properly).

wrouesnel:
This is a far simpler patch that takes advantage of the new hyperstart / runv changes. I've confirmed without it #279 is still an issue, with it, #279 is fixed.

laijs:
c.run(p) is called in the supervisor RWMutext, I would mind if it is blocked too long, otherwise I would not use go routine to create container. Is there any other solution?

wrouesnel:
Seems unavoidable to me - we can't return from the supervisor until we know the process has actually launched. Under all normal circumstances this is quick - and certainly much better behavior then now where you block indefinitely if it fails.

laijs:
you can add a 'started chan error' argument to hp.createContainer() and c.run() and "doing receive from the chan" after the lock is unlocked.

WDYT?

wrouesnel:
That would work, but it seems like it starts to complicate the design. What if instead of a global lock in `supervisor`, the supervisor lock was acquired with pod/container fidelity? That would break central point of contention in supervisor.

wrouesnel:
I looked into doing this, and it feels like a very leaky abstraction at the Supervisor level. I'm going to look into splitting the global supervisor lock into a finer-grained pod/container lock so concurrent requests can start/addProcesses to containers.

laijs:
@wrouesnel runv cli is being improved continuously and had just refactored #537 . As a result, it seams that the changes in this pr is outdated.

Close this PR. If the problem mentioned still exists, a new PR could be created for it.

PR #381, opened 2016-10-27
convert the dependency of containerd to supervisord

YaoZengzeng:
ref #381 


laijs:
keep runv-containerd for test purpose

laijs:
the runv refactor will be finished soon, we will use file-lock based management.

laijs:
@YaoZengzeng  runv cli is being improved continuously and had just refactored #537 . As a result, it seams that the changes in this pr is outdated.

PR #385, opened 2016-10-31
NewContainer: return until container is started

gao-feng:
Fix the problem that containerd send out container-start event
before container is started.


laijs:
LGTM

PR #392, opened 2016-12-07
fix dead lock for block volume insert

gnawux:
For devicemapper, introduced by refactor


gnawux:
Please take a look @laijs @gao-feng 

This fix the dead lock of block device insert

PR #399, opened 2016-12-20
fix wrong log format

resouer:
The log format is wrong when pod start successfully, encountered this when testing against `frakti`


laijs:
please push to your own tree and then create the pr

PR #401, opened 2016-12-20
Tty cleanup

laijs:


gnawux:
I really like this patch set.

PR #404, opened 2016-12-23
setup ctx of vm in associate

gao-feng:
vm.ctx should not be nil.


laijs:
it seems good in review, merge it

PR #412, opened 2017-01-06
Add a command code lookup table and adjust logging in init_comm.go

wrouesnel:
Prints the name of the command as well as the code number, rather then just the number. This greatly aids in debugging logs, since the numbers are not deterministic within the code due to iota generation.


gnawux:
I really like this feature, actually I miss it for a long time.

But I don't like to create such a lookup table manually. @wrouesnel Do you think `stringer` is better? Ref: https://godoc.org/golang.org/x/tools/cmd/stringer

wrouesnel:
Done. I've added the necessary build support to automatically pull down and run `stringer` during build with the makefile. The generated file is included in git to let a naive `go build` still work successfully.

As a side-effect, the hyperstart command codes have been broken out into their own type which considerably improves the safety of dealing with the libhyperstart code since it forces an explicit cast to functions which are handling op-codes as parameters, while still being `uint32` underneath.

PR #416, opened 2017-01-09
refactor init command handler

gao-feng:
simply logic


laijs:
could you do the refactor after libhyperstart merged? if it is still needed.....

gao-feng:
why close? i can force push this

PR #422, opened 2017-01-12
use APIVersion to wait for vm starting

gao-feng:
Change Launch function of vm as private function

### DIFF
--- a/hypervisor/vm.go
+++ b/hypervisor/vm.go
@@ -58,7 +58,7 @@ func (vm *Vm) ReleaseResponseChan(ch chan *types.VmResponse) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


laijs:
it is better to remove the waitStarted argument from GetVm(). the call sites who use waitStarted=true could use hyperstart API instead. (example: call the vm.GetHyperstartAPIVersion() instead). and some call sites who use waitStarted=true should be converted to who uses waitStarted=false.

alternative: call hyperstart API instead of WaitRespond in GetVm().

what do you think ?

gao-feng:
Will update this

gao-feng:
@laijs updated

laijs:
LGTM

PR #426, opened 2017-01-19
avoid missing event

laijs:
result := vm.WaitResponse()
send event to ctx.HUB
	[ the vm loop handles the event and sends a response event back,
	[ but vm.GetResponseChan() in the vm.WaitResponse() might be
	[ later than the response event, and the response event will not
	[ be caught, and the next statement will deadlock ]
<-result

fix it by moving vm.GetResponseChan() up (out of the go routine).


laijs:
@gao-feng

PR #427, opened 2017-01-31
Fix docker not creating stderr pipes since at least 1.13+

wrouesnel:
Docker at some point decided to stop creating stderr pipes when starting a process connected to a terminal. This patch adds some select logic to correctly support this new use case.

This problem manifests when using runv with docker versions >1.13 and possibly earlier.

This patch fixes the problem for all versions by ignoring the need to open stderr if a terminal is requested and the file does not exist. If it does exist, it is presumed the old behaviour is desired.

### DIFF
--- a/supervisor/process.go
+++ b/supervisor/process.go
@@ -38,17 +38,27 @@ func (p *Process) setupIO() error {
 		}
 	}()
 
-	stdin, err := os.OpenFile(p.Stdin, syscall.O_RDONLY, 0)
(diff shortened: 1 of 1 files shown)


gnawux:
LGTM

wrouesnel:
Actually I need to close this...while it fixes the immediate problem, it looks like I just end up not having an stderr in the terminal, so I'm not quite sure what's going on here...

PR #431, opened 2017-02-07
Fix a nil pointer dereference in supervisor/container.go:199

wrouesnel:
This use of `err` is invalid, and crashes `runv` with a nil pointer deference when err was not set in the code above. It additionally means the wrong error message is printed (as the actual check is "channel closed" not `err`).


gao-feng:
LGTM, thanks!

PR #437, opened 2017-02-14
Implement couples of APIs directly without  GenericOperation

Crazykev:
implement couples of APIs directly without  GenericOperation.

/cc @laijs PTAL

### DIFF
--- a/hypervisor/vm.go
+++ b/hypervisor/vm.go
@@ -334,10 +334,12 @@ func (vm *Vm) AddRoute() error {
 }
 
 func (vm *Vm) AddNic(info *api.InterfaceDescription) error {
+	if vm.ctx.current != StateRunning {
+		return NewNotReadyError(vm.Id)
+	}
+
 	client := make(chan api.Result, 1)
-	vm.sendGenericOperation("CreateInterface", func(ctx *VmContext, result chan<- error) {
-		go ctx.AddInterface(info, client)
-	}, StateRunning)
+	vm.ctx.AddInterface(info, client)
 
 	ev, ok := <-client
 	if !ok {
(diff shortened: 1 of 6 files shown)


hypervisor/vm.go
```diff
@@ -355,10 +357,12 @@ func (vm *Vm) AddNic(info *api.InterfaceDescription) error {
 }
 
 func (vm *Vm) DeleteNic(id string) error {
+	if vm.ctx == nil || vm.ctx.current != StateRunning {
```
laijs: after #432, we don't need to test 'vm.ctx == nil' anymore, please remove it

hypervisor/vm.go
```diff
@@ -628,30 +640,23 @@ func (vm *Vm) Pause(pause bool) error {
...
-		if ctx.PauseState == PauseStateBusy {
-			result <- fmt.Errorf("%s fails: earlier Pause or Unpause operation has not finished", command)
-			return
-		} else if ctx.PauseState == pauseState {
-			result <- nil
-			return
-		}
+	var err error
+	oldPauseState = ctx.PauseState
+	if ctx.PauseState == PauseStateBusy {
```
laijs: the old code is in statemachine, so we have PauseStateBusy for state-checking and synchronization. could you add ctx.pauseLock to protect the Pause(), and remove PauseStateBusy. (also move the changes of Pause() out from this PR, create a new one with my suggestion)

hypervisor/vm.go
Crazykev: @laijs fixed.

hypervisor/vm.go
Crazykev: @laijs I've reverted changes about Pause, will file another PR later.

laijs:
LGTM

PR #439, opened 2017-02-15
Implement vm.Pause() directly

Crazykev:
/cc @laijs As you suggested in #437, PTAL.

### DIFF
--- a/hypervisor/vm.go
+++ b/hypervisor/vm.go
@@ -627,39 +627,35 @@ func (vm *Vm) Stats() *types.PodStats {
 }
 
 func (vm *Vm) Pause(pause bool) error {
+	ctx := vm.ctx
+	if ctx.current != StateRunning {
+		return NewNotReadyError(vm.Id)
+	}
+
 	command := "Pause"
 	pauseState := PauseStatePaused
-	oldPauseState := PauseStateUnpaused
 	if !pause {
 		pauseState = PauseStateUnpaused
 		command = "Unpause"
 	}
 
-	err := vm.GenericOperation(command, func(ctx *VmContext, result chan<- error) {
-		oldPauseState = ctx.PauseState
-		if ctx.PauseState == PauseStateBusy {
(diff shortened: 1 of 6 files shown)


hypervisor/vm.go
```diff
@@ -635,30 +640,20 @@ func (vm *Vm) Pause(pause bool) error {
...
-	}, StateRunning)
-
-	if oldPauseState == pauseState {
-		return nil
+		err = ctx.DCtx.Pause(ctx, pause)
 	}
 	if err != nil {
-		pauseState = oldPauseState // recover the state
+		glog.Errorf("%s sandbox failed, recover the state", command)
+		ctx.PauseState = oldPauseState // recover the state
```
laijs: the above line and oldPauseState should be removed, ctx.PauseState has not been changed.

but you forgot to set ctx.PauseState after success.

Crazykev: Indeed, the original design lead me to expect ctx.DCtx.Pause() will change the state. Fixed and add a comment there.

laijs:
could you also fix the commit title

hypervisor/vm.go
```diff
@@ -627,39 +627,35 @@ func (vm *Vm) Stats() *types.PodStats {
...
+		return err
 	}
-	vm.GenericOperation(command+" result", func(ctx *VmContext, result chan<- error) {
-		ctx.PauseState = pauseState
-		result <- nil
-	}, StateRunning)
-	return err
+
+	glog.V(3).Infof("sandbox state turn to %s now", command)
+	ctx.PauseState = pauseState // change the state.
```
laijs: the above code should be moved up into the branch 'if ctx.PauseState != pauseState'

Crazykev:
@laijs Fixed.

laijs:
could you also use the pauseLock on Save()?

Crazykev:
@laijs Sure, will do.

PR #449, opened 2017-03-13
Fix log issues

WeiZhang555:
* Make dir before setting log dir for glog
* Don't add "\n" after glog message, because it will add \n for you
* Explicitly use glog.V(3) as debug message, and glog.V(1) as info, and
only use glog.V(1) for really important information in case it messes
all the log files. (Not finished, there's only small part here)
* Some tiny fixes


laijs:
LGTM

PR #451, opened 2017-03-15
Show commit and version numbers

WeiZhang555:
Print git commit hash id when run `runv -v`, this will be helpful for
debug.

Examples:
```
$ ./runv -v
runv version 0.7.0, commit: v0.7.0-145-g23cbab6-dirty
```

Means version: 0.7.0, latest tag is v0.7.0, we are 145 commits forward from that tag, current git commit is 23cbab6, `dirty` means we have uncommited changes in repo when building the binary.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -11,7 +11,10 @@ else
 LIBVIRT_BUILD_TAG=
 endif
 
+COMMIT=`git describe --dirty --always --tags 2> /dev/null || true`
+GOLDFLAGS="-X main.gitCommit=${COMMIT} -X main.version=${VERSION}"
 HYPER_BULD_TAGS=$(XEN_BUILD_TAG) $(LIBVIRT_BUILD_TAG)
+
 all-local: build-runv
 clean-local:
 	-rm -f runv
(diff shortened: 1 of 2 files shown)


WeiZhang555:
Another fix is, after changing the Makefile, runv is statically built binary now, before this, it's still dynamic binary.

Makefile.am
```diff
@@ -11,15 +11,19 @@ else
 LIBVIRT_BUILD_TAG=
 endif
 
+COMMIT=`git describe --dirty --always --tags 2> /dev/null || true`
+VERSION=`cat ./VERSION`
```
gao-feng: Is this ./VERSION exist?

WeiZhang555: Yes, I added a "VERSION" file in this commit.

WeiZhang555: @gao-feng Rebased.

gao-feng: @WeiZhang555 I guess you don't need VERSION file, var VERSION already exists, https://github.com/hyperhq/runv/blob/master/configure.ac#L108

WeiZhang555:
@gao-feng You're right, update now.

gao-feng:
LGTM, thanks!

PR #452, opened 2017-03-16
runv: split create/start

laijs:

### DIFF
--- a/create.go
+++ b/create.go
@@ -1,0 +1,322 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/codegangsta/cli"
+	"github.com/hyperhq/runv/containerd/api/grpc/types"
+	"github.com/kardianos/osext"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	netcontext "golang.org/x/net/context"
+)
+
+func firstExistingFile(candidates []string) string {
+	for _, file := range candidates {
+		if _, err := os.Stat(file); err == nil {
+			return file
+		}
+	}
+	return ""
+}
+
+func getDefaultBundlePath() string {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return ""
+	}
+	return cwd
+}
+
+var createCommand = cli.Command{
+	Name:  "create",
+	Usage: "create a container",
+	ArgsUsage: `<container-id>
+
+Where "<container-id>" is your name for the instance of the container that you
+are creating. The name you provide for the container instance must be unique on
+your host.`,
+	Description: `The create command creates an instance of a container for a bundle. The bundle
+is a directory with a specification file named "` + specConfig + `" and a root
+filesystem.
+
+The specification file includes an args parameter. The args parameter is used
+to specify command(s) that get run when the container is started. To change the
+command(s) that get executed on start, edit the args parameter of the spec. See
+"runv spec --help" for more explanation.`,
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "bundle, b",
+			Value: getDefaultBundlePath(),
+			Usage: "path to the root of the bundle directory, defaults to the current directory",
+		},
+		cli.StringFlag{
+			Name:  "console",
+			Usage: "specify the pty slave path for use with the container",
+		},
+		cli.StringFlag{
+			Name:  "pid-file",
+			Usage: "specify the file to write the process id to",
+		},
+		cli.BoolFlag{
+			Name:  "no-pivot",
+			Usage: "[ignore on runv] do not use pivot root to jail process inside rootfs.  This should be used whenever the rootfs is on top of a ramdisk",
+		},
+	},
+	Action: func(context *cli.Context) {
+		runContainer(context, true)
+	},
+}
+
+func runContainer(context *cli.Context, createOnly bool) {
+	root := context.GlobalString("root")
+	bundle := context.String("bundle")
+	container := context.Args().First()
+	ocffile := filepath.Join(bundle, specConfig)
+	if context.String("console") != "" {
+		fmt.Fprintf(os.Stderr, "--console is unsupported on runv\n")
+		os.Exit(-1)
+	}
+	spec, err := loadSpec(ocffile)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "load config failed: %v\n", err)
+		os.Exit(-1)
+	}
+	if spec.Linux == nil {
+		fmt.Fprintf(os.Stderr, "it is not linux container config\n")
+		os.Exit(-1)
+	}
+	if os.Geteuid() != 0 {
+		fmt.Fprintf(os.Stderr, "runv should be run as root\n")
+		os.Exit(-1)
+	}
+	if container == "" {
+		fmt.Fprintf(os.Stderr, "no container id provided\n")
+		os.Exit(-1)
+	}
+	_, err = os.Stat(filepath.Join(root, container))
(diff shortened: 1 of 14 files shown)


start.go
```diff
@@ -36,15 +36,15 @@ func getDefaultBundlePath() string {
 	return cwd
 }
 
-var startCommand = cli.Command{
-	Name:  "start",
-	Usage: "create and run a container",
+var createCommand = cli.Command{
```
WeiZhang555: Maybe we should split `createCommand` to seperate `create.go` file? :smile:

start.go
```diff
@@ -246,11 +240,106 @@ command(s) that get executed on start, edit the args parameter of the spec. See
...
+			Usage: "specify the file to write the process id to",
+		},
+		cli.BoolFlag{
+			Name:  "no-pivot",
+			Usage: "[ignore on runv] do not use pivot root to jail process inside rootfs.  This should be used whenever the rootfs is on top of a ramdisk",
+		},
+		cli.BoolFlag{
+			Name:  "detach, d",
+			Usage: "detach from the container's process",
+		},
```
WeiZhang555: Following runc command line, the `pid-file`, `no-pivot` should be flags of `create` command, `detach` should be flag of `run` command, and `console` should be replaced with `--console-socket`(we can do it later in following PR :-) ) and also it's a `create` cmd flag.

`start` command should takes no flags.

start.go
```diff
@@ -246,11 +240,106 @@ command(s) that get executed on start, edit the args parameter of the spec. See
...
+	Action: func(context *cli.Context) {
+		root := context.GlobalString("root")
+		container := context.Args().First()
+
+		if context.String("console") != "" {
+			fmt.Printf("--console is unsupported on runv\n")
+			os.Exit(-1)
+		}
+		if container == "" {
+			fmt.Printf("Please specify container ID\n")
```
WeiZhang555: Should be 
```
fmt.Fprintf(os.Stderr, "Please specify container ID\n")
```

start.go
```diff
@@ -246,11 +240,106 @@ command(s) that get executed on start, edit the args parameter of the spec. See
...
+			os.Exit(-1)
+		}
+		if os.Geteuid() != 0 {
+			fmt.Printf("runv should be run as root\n")
```
WeiZhang555: Same here.

supervisor/container.go
```diff
@@ -29,21 +29,32 @@ type Container struct {
...
+	res := c.ownerPod.vm.WaitProcess(true, []string{c.Id}, -1)
+	if res == nil {
+		c.ownerPod.sv.reap(c.Id, p.Id)
+		return fmt.Errorf("Failed to prepare waiting on the container")
+	}
 
-		err = c.start(p)
+	err = c.ownerPod.vm.Attach(p.stdio, c.Id)
+	if err != nil {
+		glog.V(1).Infof("Start Containter fail: fail to set up tty connection.\n")
```
WeiZhang555: I think it's better to use  `glog.Errorf`, and reserve `glog.V(1)` as INFO, and `glog.V(3)` as DEBUG

supervisor/container.go
```diff
@@ -114,14 +124,10 @@ func (c *Container) create(p *Process) error {
...
-
-func (c *Container) start(p *Process) error {
-
+	// save the state
 	glog.V(3).Infof("save state id %s, boundle %s", c.Id, c.BundlePath)
 	stateDir := filepath.Join(c.ownerPod.sv.StateDir, c.Id)
-	_, err := os.Stat(stateDir)
+	_, err = os.Stat(stateDir)
 	if err == nil {
 		glog.V(1).Infof("Container %s exists\n", c.Id)
```
WeiZhang555: Also here, better to use `glog.Errorf`

supervisor/supervisor.go
```diff
@@ -51,21 +51,51 @@ func New(stateDir, eventLogDir string, f factory.Factory, defaultCpus int, defau
 	return sv, sv.Events.setupEventLog(eventLogDir)
 }
 
-func (sv *Supervisor) CreateContainer(container, bundlePath, stdin, stdout, stderr string, spec *specs.Spec) (*Container, *Process, error) {
+func (sv *Supervisor) CreateContainer(container, bundlePath string, spec *specs.Spec) (c *Container, err error) {
```
WeiZhang555: As `create` command takes all the flags of original start command, I think we should save the flag configs when run the `create` command, then use `start` command to do real process starting and IO handling.

Also, I think everytime we `create` and container, `runv-containerd` should listen on a socket named with container name such as `/run/runv/containers/myContainer/containerd.sock`, so that `start` command can search the `/run/runv/containers/` dir and find the containerd socket for some container.

create.go
```diff
@@ -0,0 +1,322 @@
...
+	},
+}
+
+func runContainer(context *cli.Context, createOnly bool) {
+	root := context.GlobalString("root")
+	bundle := context.String("bundle")
+	container := context.Args().First()
+	ocffile := filepath.Join(bundle, specConfig)
+	if context.String("console") != "" {
+		fmt.Fprintf(os.Stderr, "--console is unsupported on runv\n")
```
WeiZhang555: As `--console` is unsupported flag and latest `runc` has replaced `--console` with `--console-socket` flag, I suggest we remove `--console` and add `--console-socket`, and here it will become:

```
fmt.Fprintf(os.Stderr, "--console-socket is unsupported on runv\n")
```

create.go
```diff
@@ -0,0 +1,322 @@
...
+	if context.String("pid-file") != "" {
+		stateData, err := ioutil.ReadFile(filepath.Join(context.GlobalString("root"), container, stateJson))
+		if err != nil {
+			return err
+		}
+
+		var s specs.State
+		if err := json.Unmarshal(stateData, &s); err != nil {
+			return err
+		}
```
WeiZhang555: You can move 
```
+		stateData, err := ioutil.ReadFile(filepath.Join(context.GlobalString("root"), container, stateJson))
 +		if err != nil {
 +			return err
 +		}
 +
 +		var s specs.State
 +		if err := json.Unmarshal(stateData, &s); err != nil {
 +			return err
 +		}
```

out of the `if`. Because we need to create `state.json` no matter `--pid-file` is set or not.

supervisor/container.go
WeiZhang555: Still here, we should use `glog.Errorf`

supervisor/supervisor.go
```diff
@@ -51,21 +51,46 @@ func New(stateDir, eventLogDir string, f factory.Factory, defaultCpus int, defau
...
-	return c, c.Processes["init"], nil
+	return c, nil
+}
+
+func (sv *Supervisor) StartContainer(container string, spec *specs.Spec) (c *Container, p *Process, err error) {
+	defer func() {
+		glog.V(3).Infof("Supervisor.StartContainer() return: c: %#v p: %#v", c, p)
+		if err == nil {
+			err = c.start(p)
+		}
```
WeiZhang555: This looks weird, why put the MAIN code -- `c.start(p)` -- into a `defer`?

supervisor/supervisor.go
```diff
@@ -132,19 +157,19 @@ func (sv *Supervisor) reap(container, processId string) {
 	if c, ok := sv.Containers[container]; ok {
 		if p, ok := c.Processes[processId]; ok {
 			go p.reap()
-			delete(c.ownerPod.Processes, processId)
+			delete(c.ownerPod.Processes, p.inerId)
```
WeiZhang555: small nit:   `p.innerId` ?

supervisor/supervisor.go
laijs: move it out of the lock scope

create.go
laijs: it just **reads** the state.json for the pid

PR #459, opened 2017-03-25
handle --console

laijs:


WeiZhang555:
Can't start runv container:
error log: https://gist.github.com/WeiZhang555/f8df0728bfb64c8f3afe9d69cb5be309
runv client report: error rpc error: code = Unknown desc = StartPod fail
kernel + hyper_initrd.img is fetched from hyperstart repo, commit 41262e61d1ea2ca6b768feef1bff98b568d6936f

WeiZhang555:
qemu log: https://gist.github.com/WeiZhang555/36e819b044dc6d38234b4555133c47d7

WeiZhang555:
The implementation looks good. Just wonder whether we can merge `proxy` part into `containerd` or not, or we can open a new goroutine to handle proxy function instead of adding a new `proxy` command?

PR #465, opened 2017-03-31
Support sendtty for `--console-socket` flag

WeiZhang555:
To make runv compliant with runc, we need to support `--console-socket`,
so borrow tty sending part from runc to make it behave the same.

### DIFF
--- a/shim.go
+++ b/shim.go
@@ -3,10 +3,14 @@ package main
 import (
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"syscall"
 
 	"github.com/codegangsta/cli"
+	"github.com/hyperhq/runv/containerd/api/grpc/types"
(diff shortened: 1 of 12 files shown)


shim.go
```diff
@@ -63,3 +69,33 @@ var shimCommand = cli.Command{
...
+			}
+			// forward this signal to containerd
+			sysSig, ok := s.(syscall.Signal)
+			if !ok {
+				fmt.Fprintf(os.Stderr, "can't forward unknown signal %q", s.String())
+				continue
+			}
+			if _, err := c.Signal(context.Background(), &types.SignalRequest{
+				Id:     cid,
+				Pid:    "init",
```
laijs: context.String("process")

WeiZhang555: Fixed~

PR #467, opened 2017-04-01
use hypercontaienr-utils/hlog instead of glog for some cases

gnawux:
context info will be well formatted.

And add some logs for the keep-alive check.


laijs:
LGTM

PR #475, opened 2017-04-06
*: Replace start with run and fix the description of run command

gaocegege:
start subcommand does not have "-b" flag. And it reads `state.json`, not `config.json`. It seems to be a mistake.

Signed-off-by: Ce Gao <ce.gao@outlook.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -45,7 +45,7 @@ After creating a spec for your root filesystem, you can execute a container
 in your shell by running:
 
     # cd /mycontainer
-    # runv start start [ -b bundle ] <container-id>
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -45,7 +45,7 @@ After creating a spec for your root filesystem, you can execute a container
 in your shell by running:
 
     # cd /mycontainer
-    # runv start start [ -b bundle ] <container-id>
+    # runv create [ -b bundle ] <container-id>
```
laijs: Ouch. pls replace it with `runv run`

gaocegege: Got it.

PR #478, opened 2017-04-07
interface additions: kill -a flag, ps command, pause & resume

tvelocity:
These two patches add two runc features to runv; the --all flag to kill, and the PS command. Implementing these interfaces, reveals some limitations with the current design.

Runv currently implements the OCI interface by wrapping runv-containerd calls, so we are limited by the containerd interface as supported by runv at the moment.
This introduces some limitations on these implementations, which to resolve would require the refactoring or runv, which IMHO should be discussed.

First, the processes runv is aware of from each container is limited to only the processes started directly from runv (via run and exec). Children are not being tracked. An extension in hyperstart would be needed to overcome this limitation (to always fetch the process list dynamically).

Additionally, runv-containerd reports only limited information about the processes running inside the container. We can not show the PID for example. I'm not sure whether extending hyperstart to provide this info is enough, as we need to consider how it will affect runv-containerd. Even then, we will be limited by the current fields in the containerd API. In any case runv-containerd and the runv OCI interface need to be considered at the same time to improve upon this point.

### DIFF
--- a/ps.go
+++ b/ps.go
@@ -1,0 +1,86 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"text/tabwriter"
+
+	"github.com/hyperhq/runv/containerd/api/grpc/types"
+	"github.com/urfave/cli"
+	netcontext "golang.org/x/net/context"
+)
+
+var psCommand = cli.Command{
+	Name:      "ps",
+	Usage:     "ps displays the processes running inside a container",
+	ArgsUsage: `<container-id> [ps options]`, Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "format, f",
+			Value: "table",
+			Usage: `select one of: ` + formatOptions,
+		},
+	},
+	Action: func(context *cli.Context) error {
+		container := context.Args().First()
+		if container == "" {
+			return cli.NewExitError("container id cannot be empty", -1)
+		}
+		c, err := getContainerApi(context, container)
+		if err != nil {
+			return cli.NewExitError(fmt.Sprintf("can't access container, %v", err), -1)
+		}
+
+		switch context.String("format") {
+		case "table":
+			w := tabwriter.NewWriter(os.Stdout, 12, 1, 3, ' ', 0)
+			fmt.Fprint(w, "PROCESS\tCMD\n")
+			// we are limited by the containerd interface for now
+			for _, p := range c.Processes {
+				fmt.Fprintf(w, "%s\t%s\n",
+					p.Pid,
+					p.Args)
+			}
+			if err := w.Flush(); err != nil {
+				fatal(err)
+			}
+		case "json":
+			pids := make([]string, 0)
+			for _, p := range c.Processes {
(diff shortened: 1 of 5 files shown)


tvelocity:
Added a pause & resume implementation. Same limitation as kill -a, I think an interface need to be added to hyperstart to return info about running processes to runv.

WeiZhang555:
ping @tvelocity , I think you need to do a rebase, the client lib has changed. :cat:

tvelocity:
The merge request has been updated with a rebase onto the latest master

ps.go
```diff
@@ -0,0 +1,86 @@
...
+			pids := make([]string, 0)
+			for _, p := range c.Processes {
+				pids = append(pids, p.Pid)
+			}
+
+			data, err := json.Marshal(pids)
+			if err != nil {
+				fatal(err)
+			}
+			os.Stdout.Write(data)
```
WeiZhang555: Should we print more information such as command args?

WeiZhang555:
ping @laijs @gnawux Please take a look at this :smile:

laijs:
Is the pr ready to be reviewed/merged? I will remove the `[RFC]` in the title.

ps.go
WeiZhang555: You should use `p.SystemPid` instead of `p.Pid`, it works fine according to my test

tvelocity: SystemPid reports the PID of the qemu process on the host for the container's main process, and AFAIU is invalid for the rest. See my original comment on this topic, in the first comment of this thread

> Runv currently implements the OCI interface by wrapping runv-containerd calls, so we are limited by the containerd interface as supported by runv at the moment.
> This introduces some limitations on these implementations, which to resolve would require the refactoring or runv, which IMHO should be discussed.
> 
> First, the processes runv is aware of from each container is limited to only the processes started directly from runv (via run and exec). Children are not being tracked. An extension in hyperstart would be needed to overcome this limitation (to always fetch the process list dynamically).
> 
> Additionally, runv-containerd reports only limited information about the processes running inside the container. We can not show the PID for example. I'm not sure whether extending hyperstart to provide this info is enough, as we need to consider how it will affect runv-containerd. Even then, we will be limited by the current fields in the containerd API. In any case runv-containerd and the runv OCI interface need to be considered at the same time to improve upon this point.

WeiZhang555: @tvelocity SystemPid reports PID of "nslistener" but not the qemu process, this is what we want. Because we need to make the nslistener a pseudo-container for network management. In more detail, we report "nslistener" pid as container init pid so that docker will treat "nslistener" as it's container, when CNI network plugins try to insert a network interface into the "nslistener" namespace, "nslistener" will know it and copy the network configuration into VM. This is how the network part works in runv.

PIDs in guestOS of VM is unmeaning to docker or other processes in host.

laijs: I will merge it although the problems your reported are not addressed... You can fix it later.

laijs:
Good job. thanks.

ps.go
tvelocity: @WeiZhang555 ok, thanks for the clarification

PR #480, opened 2017-04-08
Ensure the vm is running when vm operations are called

gnawux:
part of the fix of hyperhq/hyperd#598 (pod creation rollback), changes include:

- change context lock from Mutex to RWMutex
- move ctx.current access inside lock
- ensure vm operations has the running status check


laijs:
LGTM

Issue #481, opened 2017-04-09
Issue list before functional test could be merged


Crazykev:
Track issue list in runv before functional test #474 could pass and get merged.
- BUG
  - [ ]  detach mode(-d) in run/create/start works abnormally. 
- Improvement
  - [ ] #472 Link to vm root for test improvement.
  - [ ] Error message format.

gnawux:
Any update on this issue @Crazykev ?

Crazykev:
Nope, Not sure where runv should go at this point, there are still many things need to do to keep up with current containerd and OCI runtime spec.

PR #483, opened 2017-04-10
[WIP]use fifo files in non-console io mode

Crazykev:
Ref: #481 
The original design use current progress' stdio and pass them to
containerd in create stage. This will lead to `-d` flag behave wired,
since you've already passed stdin to container.

This PR enable two phase io connection use fifo pipes. In create stage,
create three named fifo files as stdio passed to container. In start
stage, open these three files and bind them with stand input/output
using `io.Copy()`

/cc @laijs 

### DIFF
--- a/create.go
+++ b/create.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


laijs:
#537 is merged, could you check if the problem is fixed please?

PR #489, opened 2017-04-13
print log after test

laijs:


Crazykev:
...Sorry, didn't expect you have done the same thing, I'll close mine later.

laijs:
the log is not printed after this patch applied, could you help and find the reason please?

Crazykev:
@laijs I guess when `cd integration-test && go test -check.v -test.timeout=120m ${TESTFLAGS} .` failed, it won't execute the next command(print the log) in Makefile

PR #490, opened 2017-04-13
Print integration test log when case failed

Crazykev:

### DIFF
--- a/integration-test/init_test.go
+++ b/integration-test/init_test.go
@@ -3,12 +3,14 @@ package main
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
 	"testing"
+	"time"
 
 	"github.com/docker/docker/integration-cli/checker"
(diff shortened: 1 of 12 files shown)


Crazykev:
@laijs PTAL
1. I've change print log logic to only print runv log when test case failed.
2. runv log always missing due to glog flush interval is 30s and can't change outside. Add a goroutine to flush glog in every 5s, and manually flush when reap pod and containerd.

integration-test/init_test.go
```diff
@@ -23,6 +25,7 @@ const (
 	initrdName     = "hyper-initrd.img"
 	binaryName     = "runv"
 	rootfsName     = "rootfs"
+	logFileName    = "runv.INFO"
```
laijs: the runv.ERROR is not printed in this patch.
if the test starts multiple runv daemon, it prints the last one only.

Crazykev: INFO has already include WARNNING and ERROR, so I skipped them.
Updated to print all the INFO log in log_dir.

laijs:
LGTM

PR #492, opened 2017-04-17
unify guest time keeping

bergwolf:
Use `-rtc base=utc,clock=vm,driftfix=slew` for both libvirt and qemu driver.


laijs:
./hypervisor/libvirt/libvirt.go:385: struct field frequency has xml tag but is not exported

PR #494, opened 2017-04-19
use acpi_pm as guest clocksource when kvm is not supported

bergwolf:
The TSC does not run at a precisely specified rate, so the guest operating system has to measure its rate at boot time, and this measurement is always somewhat inaccurate. The ACPI PM timer does run at a precisely specified rate but is slower to read than the TSC.


gnawux:
LGTM

PR #495, opened 2017-04-26
add dax support

laijs:


hypervisor/qemu/qemu.go
```diff
@@ -247,7 +247,58 @@ func (qc *QemuContext) AddDisk(ctx *hypervisor.VmContext, sourceType string, blo
...
+		}
+		size := fi.Size()
+		// compose qmp commands
+		// hmp: object_add memory-backend-file,id=mem2,share=off,mem-path=/path/to/dax.img,size=10G
+		// hmp: device_add nvdimm,id=nvdimm2,memdev=mem2
+		commands[0] = &QmpCommand{
+			Execute: "human-monitor-command",
+			Arguments: map[string]interface{}{
+				"command-line": "object_add memory-backend-file,id=mem" + strconv.Itoa(blockInfo.PmemId) +
+					",share=off,mem-path=" + filename + ",size=" + strconv.FormatInt(size, 10),
```
bergwolf: qemu docs says
```
"share=off", then guest writes won't be applied to the backend
   file and thus will be invisible to other guests.
```
Does it mean all guest modification to rootfs will be lost?

laijs:
updated.
It should be tested with 2.8 or newer qemu. (just use the official qemu).
And daxblock branch of the hyperstart.
And in /etc/hyper/config.

Hypervisor=qemu # or # Hypervisor=libvirt
StorageDriver=rawblock
StorageOptions=daxblock

laijs:
"EnableNvdimm" needs to be added to BootConfig. otherwise the PR breaks cases with the older qemus.

gnawux:
@laijs should we close this PR?

PR #496, opened 2017-04-27
clean up BlockdevInsertedEvent

bergwolf:
name, source type and scsiId are all set and passed by callers. It does
not make sense to simply pass them back.


gao-feng:
LGTM

PR #498, opened 2017-05-06
read until EOF in streamCopy

bergwolf:
If either stdoutPipe or stderrPipe reads to EOF, do not close the other
one so that it can read to EOF as well. Otherwise we might lose output
in the other pipe.

This works because if runv ends the stdout/stderr streams, it always
closes the writer part of both pipes. Then io.Copy() will get EOF and
return success here.

Also no need to wait stdin go routine because streamCopy() runs in a go
routine itself thus no one is really waiting out there.

Some observations found during looking at io.Pipe() source code:
1.   reading from a write-side closed pipe returns all buffered data and then EOF
2.   reading from a read-side closed pipe returns ErrClosedPipe and no data

So streamCopy() cannot close stdout/stderr pipes from the reader side and expect io.Copy() to still read buffered data from them.

This might fix the failure in http://ci.hypercontainer.io:8080/job/hyperd-auto/404/console

### DIFF
--- a/hypervisor/vm_states.go
+++ b/hypervisor/vm_states.go
@@ -172,6 +172,7 @@ func streamCopy(tty *TtyIO, stdinPipe io.WriteCloser, stdoutPipe, stderrPipe io.
 	var wg sync.WaitGroup
 	// old way cleanup all(expect stdinPipe) no matter what kinds of fails, TODO: change it
 	var once sync.Once
+	// cleanup closes tty.Stdin and thus terminates the first go routine
 	cleanup := func() {
 		tty.Close()
 		// stdinPipe is directly closed in the first go routine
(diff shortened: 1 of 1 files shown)


hypervisor/vm_states.go
```diff
@@ -181,30 +182,32 @@ func streamCopy(tty *TtyIO, stdinPipe io.WriteCloser, stdoutPipe, stderrPipe io.
 		}
 	}
 	if tty.Stdin != nil {
-		wg.Add(1)
```
gnawux: why remove this one?

gnawux: hmmm.... having read the commit message, looks reasonable...

gnawux:
LGTM

laijs:
LGTM

PR #509, opened 2017-05-22
remove HotAddCpuMem

laijs:
always treat it as true for x84 & arm64
always treat it as false for other arch currently.


laijs:
@gnawux @gao-feng  what do you think?

gnawux:
lgtm

PR #519, opened 2017-06-21
Don't call os.Exit explicitly to avoid lost of logs

WeiZhang555:
Do not call os.Exit directly in codes, but call "return cli.NewExitError"
instead, so that after the program terminated, glog can flush and no
logs will be lost.

### DIFF
--- a/containerd/containerd.go
+++ b/containerd/containerd.go
@@ -76,7 +76,7 @@ var ContainerdCommand = cli.Command{
 		return nil
 	},
 
-	Action: func(context *cli.Context) {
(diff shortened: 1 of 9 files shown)


WeiZhang555:
The reason that this commit works is that in https://github.com/hyperhq/runv/blob/master/main.go#L144-L148 , there is a `glog.Flush()` to flush all the logs after the APP exits, but if we call `os.Exit` in middle directly, there's no way to flush all the logs, hence cached logs would be lost. 
With this modification, `glog.Flush` will always be called when program exits normally (except panic), and we could get all logs for debugging.

PR #529, opened 2017-07-08
use machine type pc

bergwolf:
Qemu pc machine type is alias for the latest pc-i440fx-xxx series.
Use it so that we do not have to worry about updating it for different
qemu versions.


gnawux:
LGTM

bergwolf:
TODO: make it configurable, and then default to pc.

PR #530, opened 2017-07-12
Fix wrong description of runv-containerd

Hicest:
Fix miss-leading description.


gnawux:
@Hicest Thank you for the fix

PR #531, opened 2017-07-13
remove useless AllocateNetwork

gao-feng:
AllocateNetwork is replaced by ConfigureNetwork for a long time.


laijs:
It seams that ReleaseNetwork() is not called by anyone either.

gao-feng:
This should be a bug, we need release the ip address after pod destroyed at least. @laijs

laijs:
could you also fix it please?

PR #532, opened 2017-07-14
[RFC]support vhostuser for qemu driver

gao-feng:

### DIFF
--- a/hypervisor/qemu/qemu_amd64.go
+++ b/hypervisor/qemu/qemu_amd64.go
@@ -60,8 +60,13 @@ func (qc *QemuContext) arguments(ctx *hypervisor.VmContext) []string {
 		"-realtime", "mlock=off", "-no-user-config", "-nodefaults", "-no-hpet",
 		"-rtc", "base=utc,clock=vm,driftfix=slew", "-no-reboot", "-display", "none", "-boot", "strict=on",
 		"-m", memParams, "-smp", cpuParams)
-
-	if boot.BootToBeTemplate || boot.BootFromTemplate {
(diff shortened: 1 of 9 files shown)


hypervisor/qemu/qemu_amd64.go
```diff
@@ -60,8 +60,13 @@ func (qc *QemuContext) arguments(ctx *hypervisor.VmContext) []string {
 		"-realtime", "mlock=off", "-no-user-config", "-nodefaults", "-no-hpet",
 		"-rtc", "base=utc,clock=vm,driftfix=slew", "-no-reboot", "-display", "none", "-boot", "strict=on",
 		"-m", memParams, "-smp", cpuParams)
-
-	if boot.BootToBeTemplate || boot.BootFromTemplate {
+	if boot.EnableVhostUser {
```
gnawux: Does this mean the vhost-user is incompatible with template?

gao-feng: yes, vhostuser needs hugepage while template doesn't support

gnawux: any possibility to support hugepage in template in the future?

gao-feng: @laijs I think it is not easy to support hugepage for template, right?

laijs: @gao-feng  I'm not sure. And I am not even sure about whether hugepage is enabled when the memory is saved on hugetlbfs instead of tmpfs when templating.

laijs:
LGTM

PR #533, opened 2017-07-19
Add more rich network functions

WeiZhang555:
Implement more rich network functions
    
    * Support add ip addr to a specific network interface
    * Support delete ip address from a network interface
    * Support modify MTU
    * Support add any routes into container


hypervisor/network.go
```diff
@@ -282,13 +284,61 @@ func (nc *NetworkContext) getInterface(id string) *InterfaceCreated {
 	return nil
 }
 
+func (nc *NetworkContext) AddIPAddr(id, ip string) error {
+	nc.slotLock.RLock()
```
gao-feng: Should not be read lock here.

hypervisor/network.go
```diff
@@ -282,13 +284,61 @@ func (nc *NetworkContext) getInterface(id string) *InterfaceCreated {
...
+		return ErrNoSuchInf
+	}
+	inf.IpAddr = append(inf.IpAddr, ip)
+	inf.Desc.Ip = append(inf.Desc.Ip, ip)
+	return nil
+}
+
+func (nc *NetworkContext) DeleteIPAddr(id, ip string) error {
+	nc.slotLock.RLock()
+	defer nc.slotLock.RUnlock()
```
gao-feng: same here

hypervisor/network.go
```diff
@@ -282,13 +284,61 @@ func (nc *NetworkContext) getInterface(id string) *InterfaceCreated {
...
+	inf, ok := nc.idMap[id]
+	if !ok {
+		return ErrNoSuchInf
+	}
+
+	for i, v := range inf.IpAddr {
+		if v == ip {
+			inf.IpAddr = append(inf.IpAddr[:i], inf.IpAddr[i+1:]...)
```
gao-feng: what happen when the last element in ipaddr array match the ip, is the inf.IpAddr[i+1] will access the array out of range?

hypervisor/network.go
```diff
@@ -282,13 +284,61 @@ func (nc *NetworkContext) getInterface(id string) *InterfaceCreated {
...
+			break
+		}
+	}
+
+	for i, v := range inf.Desc.Ip {
+		if v == ip {
+			inf.IpAddr = append(inf.Desc.Ip[:i], inf.Desc.Ip[i+1:]...)
```
gao-feng: same here

hypervisor/network.go
```diff
@@ -282,13 +284,61 @@ func (nc *NetworkContext) getInterface(id string) *InterfaceCreated {
...
+			break
+		}
+	}
+	return nil
```
gao-feng: return error or printout warning info if there is no matching ip?

hypervisor/network.go
```diff
@@ -282,13 +284,61 @@ func (nc *NetworkContext) getInterface(id string) *InterfaceCreated {
...
+}
+
+func (nc *NetworkContext) UpdateMtu(id string, mtu uint64) error {
+	nc.slotLock.RLock()
+	defer nc.slotLock.RUnlock()
```
gao-feng: Seems like write lock is more appropriate here

hypervisor/network/network.go
```diff
@@ -9,19 +9,20 @@ import (
...
+	Bridge    string
+	Device    string
+	File      *os.File
+	Automatic bool
 }
 
 const (
 	DefaultBridgeIface = "hyper0"
 	DefaultBridgeIP    = "192.168.123.0/24"
+	DefaultMtu         = 1450
```
gao-feng: why 1450 by default? it should not be 1500?

hypervisor/network/network_linux.go
```diff
@@ -1049,45 +1049,27 @@ func AllocateAddr(requestedIP string) (*Settings, error) {
...
-		IPAddress:   ip.String(),
-		Gateway:     BridgeIPv4Net.IP.String(),
-		Bridge:      BridgeIface,
-		IPPrefixLen: maskSize,
-		Device:      "",
-		File:        nil,
-		Automatic:   true,
+		Mac:       mac,
+		IPAddress: []string{ip.String()},
+		Mtu:       DefaultMtu,
```
gao-feng: I think by default we should not care about mtu, vm/os will setup the appropriate mtu for us. lets just set it null.

hypervisor/vm_states.go
```diff
@@ -106,11 +108,85 @@ func (ctx *VmContext) restoreContainer(id string) (alive bool, err error) {
...
+		}
+		return nil
+	}
+}
+
+func (ctx *VmContext) hyperstartAddRoute(r []hyperstartapi.Route) error {
+	if err := ctx.hyperstart.AddRoute(r); err != nil {
+		return err
+	}
+	return nil
```
gao-feng: return ctx.hyperstart.AddRoute(r)?

supervisor/proxy/nslistener.go
```diff
@@ -213,19 +212,38 @@ func handleAddr(update netlink.AddrUpdate, callback func(supervisor.NetlinkUpdat
...
 
 	netlinkUpdate := supervisor.NetlinkUpdate{}
 	netlinkUpdate.Route = update
 	netlinkUpdate.UpdateType = supervisor.UpdateTypeRoute
 	callback(netlinkUpdate)
+
+	// delete all the routes from outside
+	// The reason that we need to delete all the routes is
+	// that if we don't do this, we will get two active network interfaces,
+	// and ping the IP will get duplicate ICMP replies.
```
gao-feng: The nic in netns of docker container(netns of nslistener) should be down, is it still has this problem?

hypervisor/network.go
WeiZhang555: You're right.

hypervisor/network.go
WeiZhang555: It should be fine, golang will handle it. See https://play.golang.org/p/5SWxZB3p2r

hypervisor/network/network.go
WeiZhang555: I remember that vxlan interface has a mtu of 1450, once I met a problem that container network can't visit outside via vxlan, after setting its mtu to 1450, everything is good again.

But I'll keep mtu empty to let os handle it as your other comments suggest.

hypervisor/vm_states.go
WeiZhang555: Right :+1:

supervisor/proxy/nslistener.go
WeiZhang555: I think we can still enter the netns and make the nic up? This should be a quite easy step, so we can't guarantee CNI plugin won't set nic up later.

Delete all route should be another protection for duplicate network flows.

gao-feng: I'm wonder if this deleting of router of netns will trigger route-update netlink event, though we don't care about this event right now.

Maybe removing the veth pair on host which joins the docker bridge is a good way, but the code will be complicated...

WeiZhang555: > Maybe removing the veth pair on host which joins the docker bridge is a good way

This is also another option, maybe we should do this?  For example, after tap created and inserted, removed veth from bridge?

gnawux:
@WeiZhang555 could you help do a small rebase? The #537 introduced some conflicts

WeiZhang555:
@gnawux I'll do it soon :wink:

supervisor/proxy/nslistener.go
WeiZhang555: @gao-feng I just met a extremely weird problem, starting two container in one VM(POD), first one has proper route rules, but second container has no route rules. How could this happen? I thought they were in same network namespace and shares all routes.  Do you have any idea?

gao-feng: @WeiZhang555 
Can you check the ipaddress of nic? is it same on two container? they should stay in the same netns.

WeiZhang555: Yes, the nic ip is the same, only routes are different, quite weird. I'll try to do more investigation.

WeiZhang555:
Closing this. We have a new one #597 carrying this :smile:

PR #535, opened 2017-07-27
Fixes #527 that mistook "init" for process name

### LINKED ISSUE #527: `docker exec` will stuck when exit
Start a runv container with docker, then `docker exec -ti xxx sh`, after typing `exit` in `sh`, the client side will stuck. I have to use `kill <docker exec pid>` to stop the client.

I was trying to debug this, find that the `docker-containerd-shim` is stuck on waiting IO terminating. then I checked runv codes, and didn't find the problem, the IO channel seems closed correctly. 

I've no idea how to fix this, so file an issue here. :)

yyb196:
Fixes #527 that mistook "init" for process name
to monitor in case of exec command, which makes
exec command whose process has exited stuck

Signed-off-by: frank yang <yyb196@gmail.com>

### DIFF
--- a/create.go
+++ b/create.go
@@ -245,7 +245,7 @@ func createContainer(context *cli.Context, container, namespace string, config *
 		return err
 	}
 
-	return ociCreate(context, container, func(stdin, stdout, stderr string) error {
+	return ociCreate(context, container, "init", func(stdin, stdout, stderr string) error {
 		r := &types.CreateContainerRequest{
 			Id:         container,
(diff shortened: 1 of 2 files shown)


laijs:
LGTM

WeiZhang555:
I'm sorry to say that this doesn't fix #527, in my test, `exec` still stuck, though this patch is doing the right thing.

We need to reopen that issue for tracking.

image

PR #539, opened 2017-08-07
move iptables related code from runv to hyperd

gnawux:
runv does not setup port mapping, which is a hyperd daemon function.
Let's move it back to make things easier.


gnawux:
Based on the discussion this morning, this pr is updated, which moves iptables related code out from runv.

And the code is moved to hyperd via hyperhq/hyperd#652

gnawux:
This PR could be considered as a clean up. Any comments @gao-feng ?

gao-feng:
LGTM

PR #543, opened 2017-08-12
support xen-pv

gao-feng:
Env:
xen: 4.9
qemu: 2.10.0
guest kernel: 4.12.4 

Todo:
a: Use one xenlight library for xen pv and hvm.
b: use vm_console to handle pv domain console output (done)
c: fix some weird problems in some wrong case.
1, host dom name will be (null) and fail to create xenpv domain(fixed)
2,hyperd panic when fail to create xenpv domain (fixed)
3,hyperd hang (disappear after fix bug 2)

### DIFF
--- a/hypervisor/vm_console.go
+++ b/hypervisor/vm_console.go
@@ -59,24 +59,28 @@ func watchVmConsole(ctx *VmContext) {
 		return
 	}
 
-	conn, err := utils.UnixSocketConnect(ctx.ConsoleSockName)
-	if err != nil {
-		ctx.Log(ERROR, "failed to connected to %s: %v", ctx.ConsoleSockName, err)
-		return
-	}
-
-	ctx.Log(TRACE, "connected to %s", ctx.ConsoleSockName)
-
-	tc, err := telnet.NewConn(conn)
-	if err != nil {
-		ctx.Log(ERROR, "fail to init telnet connection to %s: %v", ctx.ConsoleSockName, err)
-		return
-	}
-	ctx.Log(TRACE, "connected %s as telnet mode.", ctx.ConsoleSockName)
-
 	cout := make(chan string, 128)
-	go TtyLiner(tc, cout)
+	if dctx, ok := ctx.DCtx.(ConsoleDriverContext); ok {
+		dctx.ConnectConsole(cout)
+	} else {
 
+		conn, err := utils.UnixSocketConnect(ctx.ConsoleSockName)
+		if err != nil {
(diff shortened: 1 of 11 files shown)


hypervisor/xenpv/xenpv.go
```diff
@@ -0,0 +1,285 @@
...
+
+	xc.connectConsole(ctx.Id, xc.domId)
+	go func() {
+		//unpause dom until runv connected to the socket, otherwise may loss the ready message
+		ctx.WaitSockConnected()
+		xctx.DomainUnpause(xc.domId)
+	}()
+}
+
+func (xc *XenPvContext) connectConsole(vmid string, domId xl.Domid) error {
```
laijs: Could it be possible to merge it into vm_console.go?

gao-feng: I can only add another unix socket to pass console data, seems like pv doesn't have direct way to get console from unix socket.

laijs: It seems that adding `ConnectConsole() io.Reader`  to the driver interface is a possible way to merge the code. What do you think?

gao-feng: @laijs 
Good point!
Commit two adds ConsoleReader to the driver interface,  please check.

hypervisor/vm_console.go
```diff
@@ -59,20 +59,19 @@ func watchVmConsole(ctx *VmContext) {
...
-		ctx.Log(ERROR, "failed to connected to %s: %v", ctx.ConsoleSockName, err)
+		ctx.Log(ERROR, "failed to get console for %v %v", ctx.Id, err)
 		return
 	}
+	defer reader.Close()
 
-	ctx.Log(TRACE, "connected to %s", ctx.ConsoleSockName)
-
-	tc, err := telnet.NewConn(conn)
+	tc, err := telnet.NewConn(reader)
```
laijs: It seams that the above code(L62-L69) should all be moved in to DCtx.ConsoleReader() (those who are using ctx.ConsoleSockName).
DCtx.ConsoleReader() for xenpv doesn't returns a telnet socket, you cannot use telnet.NewConn() for it.

gao-feng:
@laijs 
updated, xenlight-runv.go needs to access the private data `Context.ctx` in xenlight package, looks like use xenlight package as vendor is impossible.

gnawux:
hmm.... looks we need submit change request to official xenlight go binding in the future

PR #544, opened 2017-08-14
track and report container state correctly

bergwolf:
containerd-shim will query container status before deleting it.


cli/container.go
```diff
@@ -110,6 +118,7 @@ func createContainer(options runvOptions, vm *hypervisor.Vm, container, bundle,
 			Pid:     shim.Pid,
 			Bundle:  bundle,
 		},
+		stat.StartTime, 0,
```
laijs: please use ShimCreateTime: stat.StartTime

bergwolf: golang does not allow mixing field:value and value initializers. e.g.,
`cli/container.go:114: mixture of field:value and value initializers`

bergwolf: You were right. I've fixed it.

laijs:
LGTM

PR #545, opened 2017-08-14
return success if the container does not exist when force

laijs:

### DIFF
--- a/cli/delete.go
+++ b/cli/delete.go
@@ -44,6 +44,17 @@ status of "ubuntu01" as "stopped" the following will delete resources held for
 
 		state, spec, err := loadStateAndSpec(root, container)
 		if err != nil {
+			// return success if the container does not exist when force
+			// and also try to remove the empty container state dir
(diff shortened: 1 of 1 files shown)


cli/delete.go
```diff
@@ -44,6 +44,17 @@ status of "ubuntu01" as "stopped" the following will delete resources held for
 
 		state, spec, err := loadStateAndSpec(root, container)
 		if err != nil {
+			// return success if the container does not exist when force
+			// and also try to remove the empty container state dir
+			if force {
+				if errRmdir := os.Remove(filepath.Join(root, container)); errRmdir == nil || os.IsNotExist(errRmdir) {
+					return nil
+				}
+				_, errState := os.Stat(filepath.Join(root, container, stateJSON))
```
bergwolf: remove state.json before removing the state dir? os.Remove fails if the directory is not empty.

laijs: if state.json exists, just return error

PR #554, opened 2017-08-14
kill two todos in deleteContainer operation

bergwolf:
Also clean up glog trailing newline a bit in cli/container.go.

### DIFF
--- a/cli/container.go
+++ b/cli/container.go
@@ -145,11 +145,14 @@ func createContainer(options runvOptions, vm *hypervisor.Vm, container, bundle, 
 }
 
 func deleteContainer(vm *hypervisor.Vm, root, container string, force bool, spec *specs.Spec, state *State) error {
-
-	// todo: check the container from vm.ContainerList()
-
 	// non-force killing can only be performed when at least one of the realProcess and shimProcess exited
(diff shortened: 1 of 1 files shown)


cli/container.go
```diff
@@ -145,13 +145,18 @@ func createContainer(options runvOptions, vm *hypervisor.Vm, container, bundle,
...
-	exitedHost := syscall.Kill(state.Pid, syscall.Signal(0)) != nil
+	exitedVM := true
+	for _, c := range vm.ContainerList() {
+		if c == container {
+			exitedVM = vm.SignalProcess(container, "init", syscall.Signal(0)) != nil // todo: is this check reliable?
+			break
+		}
+	}
+	exitedHost := true
+	if containerShimAlive(state) {
```
laijs: Is system.Stat() more reliable than syscall.Kill() in which case we don't need syscall.Kill() anymore?

bergwolf: Good point! Let's replace syscall.Kill() with system.Stat() for shim process checking.

bergwolf:
@laijs updated.

laijs:
LGTM

Issue #550, opened 2017-08-14
rootfs in the bundle may be corrupted in some cases.


laijs:
result: many files/dirs in the rootfs in the bundle was removed when failed
cases: unknown...
reason(probably): the sharedir was failed to be umounted, and the files would be removed when runv was removing the vm(and its related dirs).

PR #557, opened 2017-08-15
reorder createContainer steps

bergwolf:
defer functions are called in stack order. Move state dir preparation
first so that it is cleaned up last. Otherwise we might accidentally
delete contents in container rootfs in case of failure.

This should fix https://github.com/hyperhq/runv/issues/550.


bergwolf:
os.RemoveAll() does not follow symlinks!

PR #562, opened 2017-08-18
update runv state to match OCI definition

bergwolf:
Per
https://github.com/opencontainers/runtime-spec/blob/master/runtime.md,
`runv state` should return something like:

```
{
    "ociVersion": "0.2.0",
    "id": "oci-container1",
    "status": "running",
    "pid": 4422,
    "bundle": "/containers/redis",
    "annotations": {
        "myKey": "myValue"
    }
}
```

So we return:
```
$sudo runv --root /run/containerd/runc/default state foobar
{
  "ociVersion": "1.0.0",
  "id": "foobar",
  "pid": 95193,
  "bundle": "/run/containerd/io.containerd.runtime.v1.linux/default/foobar",
  "status": "running",
  "annotations": null
}
```

### DIFF
--- a/cli/state.go
+++ b/cli/state.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"time"
 
 	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/urfave/cli"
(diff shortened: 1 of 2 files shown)


bergwolf:
dropped.

PR #566, opened 2017-08-21
add owner field to list and state cmd

bergwolf:
To match runc outputs.

Also remove ps options and runv does not support it.

And add a default error code in case runv cli fails but no error is set.


cli/main.go
```diff
@@ -146,6 +146,9 @@ func main() {
 	if err := app.Run(os.Args); err != nil {
 		glog.Errorf("app.Run(os.Args) failed with err: %#v", err)
 		fmt.Fprintf(os.Stderr, "%v\n", err)
+		cli.HandleExitCoder(err)
+		// non-standard errors
+		os.Exit(22)
```
laijs: what if err == nil?

laijs:
LGTM

PR #572, opened 2017-08-24
Enable pod support

ZhaoXuqiang:
share a pod with existing container, with `--net container:xxx`
get shared container by compare pid with exist running container's Pid which found from container's state file

Signed-off-by: ZhaoXuqiang <zhaoxuqiang@huawei.com>

### DIFF
--- a/cli/create.go
+++ b/cli/create.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/hyperhq/runv/hypervisor"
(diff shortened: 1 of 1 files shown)


WeiZhang555:
Please modify your commit message to describe more about this modifcations, also remove "cherry-pick from cfd4695 and 81d42e4", 81d42e4 isn't included in these codes and is outdated so you don't need to mention it. @ZhaoXuqiang

cli/create.go
```diff
@@ -95,20 +97,11 @@ func cmdCreateContainer(context *cli.Context, attach bool) error {
...
-				sharedContainer = ns.Path
-				_, err = os.Stat(filepath.Join(root, sharedContainer, stateJSON))
-				if err != nil {
-					return fmt.Errorf("The container %q is not existing or not ready", sharedContainer)
-				}
-				_, err = os.Stat(filepath.Join(root, sharedContainer, "namespace"))
-				if err != nil {
-					return fmt.Errorf("The container %q is not ready", sharedContainer)
+				if sharedContainer, err = findSharedContainer(context.GlobalString("root"), ns.Path); err != nil {
+					return fmt.Errorf("failed to find shared container: %v", err)
```
WeiZhang555: Can you keep 
```
-				_, err = os.Stat(filepath.Join(root, sharedContainer, stateJSON))
 -				if err != nil {
 -					return fmt.Errorf("The container %q is not existing or not ready", sharedContainer)
 -				}
 -				_, err = os.Stat(filepath.Join(root, sharedContainer, "namespace"))
 -				if err != nil {
 -					return fmt.Errorf("The container %q is not ready", sharedContainer)
```

This piece of code is used for validation, I think it's still useful.

ZhaoXuqiang: it's not useful.
```
 	if sharedContainer != "" {
 		scState, err = loadStateFile(root, sharedContainer)
 		if err != nil {
 			return err
 		}
 		vm, lockFile, err = getSandbox(filepath.Join(context.GlobalString("root"), sharedContainer, "sandbox"))
 		if err != nil {
 			return err
 		}
 	}
```
```
func loadStateFile(root, container string) (*State, error) {
	stateFile := filepath.Join(root, container, stateJSON)
	file, err := os.Open(stateFile)
	if err != nil {
		return nil, err
	}
	defer file.Close()

	var state State
	if err = json.NewDecoder(file).Decode(&state); err != nil {
		return nil, fmt.Errorf("Decode state file %s error: %s", stateFile, err.Error())
	}
	return &state, nil
}
```
loadStateFile will confirm the state file exist.
namespace file is not used.

WeiZhang555: Looks like it's OK then.

ping @laijs

laijs:
LGTM

PR #575, opened 2017-08-24
Network releated update

gao-feng:
1,  BuildinNetworkDriver represents the hypervisor driver has self suitable network model instead of general bridge/port, currently only vbox dirver implements BuildinNetworkDriver. (Is vbox dirver useful?)

2, Move the codes that creating tap device from network package to qemu driver, Time has proved only qemu needs this。

3, use netlink go package to simplify runv network codes

4, some cleanup


gnawux:
LGTM

I don't think vbox driver is useful now. And does vbox driver work currently?

laijs:
It's very good!

PR #587, opened 2017-08-30
hyperstart: cache the stream before consumed

laijs:


lib/utils/bytespipe.go
```diff
@@ -0,0 +1,188 @@
+package utils
+
+// copied from moby
```
bergwolf: ditto here.

lib/utils/buffer.go
```diff
@@ -0,0 +1,53 @@
+package utils
+
+// copied from moby
```
bergwolf: Why not simply import "github.com/moby/moby/pkg/ioutils"?

bergwolf:
After talking to @laijs, we copy the files from moby since runv already has imports from docker.

Issue #586, opened 2017-08-30
Support set vcpu number


gnawux:
The hypervisor driver package support configure vCPU number([code here](https://github.com/hyperhq/runv/blob/master/hypervisor/driver.go#L13)), however, we could not setup it through runv command line.

As [oci spec](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cpu) does not specified it. There are several alternatives:

- use `cpu.period` and `cpu.quota` (also `docker --cpus`), and do a round up for float number.
- use `cpu.cpus` (i.e. `docker --cpuset`)
- use a self defined environment

@jovizhangwei suggest we could use the `docker --cpus` value because it works well with many 3rd party control console (such as Docker UCP).

PR #589, opened 2017-09-01
respect cpu and memory settings in spec

bergwolf:
For CPU, use Spec.Linux.Resources.CPU.Quota/Spec.Linux.Resources.CPU.Period.

For Memory, use Spec.Linux.Resources.Memory.Limit.
```
$docker run --rm -it --cpus 1.5 --memory 671088640 busybox sh
/ # cat /proc/cpuinfo |grep processor
processor       : 0
processor       : 1
/ # cat /proc/meminfo |grep MemTotal
MemTotal:         569512 kB
```

fixes https://github.com/hyperhq/runv/issues/586

### DIFF
--- a/cli/create.go
+++ b/cli/create.go
@@ -124,7 +124,8 @@ func cmdCreateContainer(context *cli.Context, attach bool) error {
 		if err != nil {
 			return nil
 		}
-		vm, lockFile, err = createAndLockSandBox(f, spec, context.GlobalInt("default_cpus"), context.GlobalInt("default_memory"))
+		cpu, mem := getContainerCPUMemory(context, spec)
+		vm, lockFile, err = createAndLockSandBox(f, spec, cpu, mem)
 		if err != nil {
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


cli/create.go
```diff
@@ -140,6 +141,32 @@ func cmdCreateContainer(context *cli.Context, attach bool) error {
...
+		resource := spec.Linux.Resources
+		if resource.CPU != nil && resource.CPU.Period != nil && resource.CPU.Quota != nil {
+			period := *resource.CPU.Period
+			quota := *resource.CPU.Quota
+			if period > 0 && quota > 0 {
+				cpu = int((uint64(quota) + period - 1) / period)
+			}
+		}
+		if resource.Memory != nil && resource.Memory.Limit != nil {
+			mem = int(*resource.Memory.Limit / 1024 / 1024)
```
laijs: createAndLockSandBox() has the same code, I think is should be removed.

bergwolf:
@laijs updated.

laijs:
LGTM

PR #590, opened 2017-09-01
return cleanly when return non-zero

laijs:
remove the incorrectly appended "<nil>":

runv exec test_busybox command-does-not-exist
exec failed: No such file or directory
<nil>


bergwolf:
Should we fix osProcessWait() instead? e.g., make sure it always set err if exit status is non-zero.

laijs:
non-zero exitcode returns from the process doesn't mean error.

laijs:
@bergwolf updated

Issue #582, opened 2017-08-28
Test runv failed in swarm mode


jovizhangwei:
I've tested runv in swarm mode, failed with below logs:

```
$sudo docker service create --name nginx -p 8080:80 nginx

$sudo docker service logs nginx
nginx.1.jvu6senhvj68@buffer010101170038.et2sqa    | E0828 16:55:30.287532   31315 network.go:153] cann't find nic whose ifindex is 9210
nginx.1.jvu6senhvj68@buffer010101170038.et2sqa    | E0828 16:55:30.288391   31315 network.go:153] cann't find nic whose ifindex is 9210
```

Instead of swarm mode, nginx could started normally by "docker run".
```
$sudo docker run -it --rm -p 8081:80 nginx bash
root@702b784a1666:/# nginx
root@702b784a1666:/# 
```

In contrary, Clear Container runs smoothly in swarm mode.

gao-feng:
Hi

Can you checkout if this pr fix this problem?
https://github.com/hyperhq/runv/pull/592

Thanks!

jovizhangwei:
Thanks for your work @gao-feng , does that PR already supported below commands?

```
docker service create ...
docker service create --network overlay-net ...
```

gao-feng:
@jovizhangwei 

Yes, with my test, these commands works well, but FYI, runv doesn't support DNS server in user-defined networks now https://github.com/hyperhq/runv/issues/363

Thanks

jovizhangwei:
I can ran below command:
```
$sudo docker service create --name nginx -p 8080:80 nginx:alpine
xbh2i3cs23ntwsnnyq7fuugvb
Since --detach=false was not specified, tasks will be created in the background.
In a future release, --detach=false will become the default.
```
But I cannot access the nginx service from outside, I can confirm the nginx can receive the request, but some issues happened when send out the request.

Everything is fine if I start the container not it swarm mode.
```
$sudo docker run -it --rm -p 8080:80 nginx:alpine sh
/ # nginx 
```

gao-feng:
test nginx:alpine image on swarm mode, it works well, I can access nginx Welcome page on 172.18.0.x address, did you use ping to test if the network is available?

jovizhangwei:
@gao-feng  Still not working, same error as last post.

The network is available, and the nginx can be accessed from outside if started by "docker run", but can not access if started by "docker  service create", both use runv as runtime.

Run command both in runv and Clear Container:
```
[...]

PR #592, opened 2017-09-05
use tc mirred rule to connect hypervisor nic and container nic

gao-feng:
Try to fix https://github.com/hyperhq/runv/issues/582

this pr adds tc mirred rules in the netns of shim process to connect two type of nics together without regarding for the docker network mode.

macvtap may be an alternate, but it can only be used on qemu(xen?). 

need more test.


lib/runvxenlight/xenlight-runv.go
```diff
@@ -124,14 +124,15 @@ void runvxl_childproc_setmode(libxl_ctx *ctx) {
...
-int runvxl_add_nic(libxl_ctx *ctx, int32_t domid, int id, char *bridge, char *mac) {
+int runvxl_add_nic(libxl_ctx *ctx, int32_t domid, int id, char *bridge, char *device, char *mac) {
     libxl_device_nic nic;
     int ret;
 
     libxl_device_nic_init(&nic);
     nic.bridge = strdup(bridge);
     nic.nictype = LIBXL_NIC_TYPE_VIF;
     nic.devid = id;
+    nic.ifname = strdup(device);
```
laijs: where is the code that nic.ifname and nic.bridge are to be freed?
Is it libxl_device_nic_dispose(&nic); ?

cli/network.go
```diff
@@ -53,68 +53,14 @@ type nsListener struct {
...
+func createFakeBridge() {
+	// add an useless bridge to satisfy hypervisor, most of them need to join bridge.
+	la := netlink.NewLinkAttrs()
+	la.Name = fakeBridge
+	bridge := &netlink.Bridge{LinkAttrs: la}
+	if err := netlink.LinkAdd(bridge); err != nil && !os.IsExist(err) {
+		glog.Errorf("fail to create fake bridge: %v, %v", fakeBridge, err)
 	}
-
-	if bridge.Name == "ovs-system" {
```
laijs: Why is this part removed? Is this code also a trick?

cli/network.go
```diff
@@ -25,6 +24,7 @@ const (
 	UpdateTypeLink  NetlinkUpdateType = "link"
 	UpdateTypeAddr  NetlinkUpdateType = "addr"
 	UpdateTypeRoute NetlinkUpdateType = "route"
+	fakeBridge      string            = "runv0"
```
laijs: all taps will be added to this bridge. I'm afraid it might cause information leak.

laijs: when the taps are moved into shim's ns, they are also atomically moved out from the bridge. But what happen before the moving?

lib/runvxenlight/xenlight-runv.go
gao-feng: yes

cli/network.go
gao-feng: Nop, the new design doesn't need to lookup bridge/ovs on host now

cli/network.go
gao-feng: Have no idea, What's your opinion?

laijs:
LGTM

PR #593, opened 2017-09-06
return error when initSandboxNetwork fail

Ace-Tang:
deal error when initSandboxNetwork() fail

Signed-off-by: Ace-Tang <aceapril@126.com>


Ace-Tang:
I get error when using runv as a runtime,  it looks successful in create process, but fail at start container process, at last I found initSandboxNetwork() fail,  but startNsListener() not deal with it, so I wonder if deal with error is need

Ace-Tang:
seems initSandboxNetwork return error all the time,

Issue #591, opened 2017-09-04
Proposal: support CNI network plugins


WeiZhang555:
Currently runv master can't operate with CNI network plugins, I think we should add some network operation interfaces for runv to make it work with CNI. Interfaces could be like this:

```
Commands:
	interface	manage network interface
	route		manage route
```

## 1. runv interface

```
Subcommands:
	add	add an interface into a container
	rm	remove an interface from a container
	update	update configuration of an interface
	ls	list network interfaces in an container
	inspect show configuration of a specific network interface from a container
```

### 1.1. add
```
usage: add container
flags:
	--tapname	use existing tap device in host, tap will be created for you if empty
	--name		set interface name in container
	--ip		set ip address with a mask, format: 192.168.0.2/24
	--mac		set mac address 
	--mtu		set mtu
```
	
### 1.2. rm
```
Usage: rm container name - rename interface "name" from container "container"
flags:
```
	
### 1.3 update
```
Usage: update container interface - update configuration of interface "interface" in a container
flags:
	--name		set interface name in container
	--add-ip	add one more ip address with a mask, format: 192.168.0.2/24
	--rm-ip		remove one ip address from a container, format: 192.168.0.2/24
	--mac		set mac address 
	--mtu		set mtu
```

### 1.4 ls
```
Usage: ls container - list all interfaces in a container
```

### 1.5 inspect
```
Usage: inspect container interface - inspect interface configuration in a container
```
	
### 2.1 add
```
Usage: add container route - add route rule into a container
flags:
	--dest		the destination prefix of the route
	--dev		the output device name
	--src		the source address to prefer when sending to the destinations covered by the route prefix.
	--gateway	the address of the nexthop router
```

### 2.2 rm
```
Usage: remove container route - remove route rule from container, "route" is destination prefix of the route
```
	
### 2.3 ls
```
Usage: ls container - list all route rules in a container
```

PR #597, opened 2017-09-12
Enhance network functions

WeiZhang555:
Fixes part of https://github.com/hyperhq/runv/issues/591
Carry #533 

Mainly add "interface" command and some sub-commands to implement a first-class network interface operation.

Note that "remove interface" command rely on hyperstart PR: https://github.com/hyperhq/hyperstart/pull/318

Sorry that this hasn't been rebased to latest master, but I think it will still be useful.


cli/interface.go
```diff
@@ -0,0 +1,215 @@
...
+	Name:  "interface",
+	Usage: "manage interfaces for container",
+	Subcommands: []cli.Command{
+		infAddCommand,
+		infRmCommand,
+		infUpdateCommand,
+		infListCommand,
+	},
+	Before: func(context *cli.Context) error {
+		return cmdPrepare(context, true, context.Bool("detach"))
```
gao-feng: Should context.Bool("detach")  be changed to true?

cli/interface.go
```diff
@@ -0,0 +1,215 @@
...
+	},
+	Action: func(context *cli.Context) error {
+		return nil
```
gao-feng: Action function should be cli.ShowSubcommandHelp(context)?

cli/interface.go
```diff
@@ -0,0 +1,215 @@
...
+	},
+}
+
+var infAddCommand = cli.Command{
+	Name:      "add",
+	Usage:     "add an interface into a container",
+	ArgsUsage: `add <container-id>`,
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "tapname",
```
gao-feng: I think ```hostdev``` may be general name, ```tapname``` is too specialized name.

cli/interface.go
```diff
@@ -0,0 +1,215 @@
...
+	},
+	Action: func(context *cli.Context) error {
+		container := context.Args().First()
+		vm, releaseFunc, err := vmByContainerID(context, container)
+		if err != nil {
+			return err
+		}
+		defer releaseFunc()
+
+		ip := context.String("ip")
```
gao-feng: why getting ip separately here?

cli/interface.go
```diff
@@ -0,0 +1,215 @@
...
+		defer releaseFunc()
+
+		targetInf := context.Args().Get(1)
+		if targetInf == "" {
+			return cli.NewExitError("please specify an interface name to update", -1)
+		}
+
+		conf := &api.InterfaceDescription{
+			Id:   "-1",
+			Name: context.String("name"),
```
gao-feng: What's the value of context.String("name")? is it existed?

cli/network.go
```diff
@@ -110,6 +113,9 @@ func initSandboxNetwork(vm *hypervisor.Vm, enc *gob.Encoder, dec *gob.Decoder, p
 			Lo:     false,
 			Bridge: fakeBridge,
 			Ip:     info.Ip,
+			Name:   info.Name,
+			Mac:    info.Mac,
```
gao-feng: So nic in vm and nic in netns will share the same mac address, I think this is ok, can you check if this will cause some problem in network test?

hypervisor/network.go
```diff
@@ -216,18 +308,18 @@ func (nc *NetworkContext) netdevInsertFailed(idx int, name string) {
 
 func (nc *NetworkContext) configureInterface(index, pciAddr int, name string, inf *api.InterfaceDescription, result chan<- VmEvent) {
 	if inf.TapName == "" {
-		inf.TapName = network.NicName(nc.sandbox.Id, index)
```
gao-feng: If you want to change the tapname, please update network.NicName function, other place rely on this naming rule.

hypervisor/network/network_linux.go
```diff
@@ -201,14 +201,16 @@ func UpAndAddToBridge(name, bridge, options string) error {
...
 	}
-	master, err := netlink.LinkByName(bridge)
-	if err != nil {
-		glog.Error("cannot find bridge interface ", bridge)
-		return err
-	}
-	if err = addToBridge(iface, master, options); err != nil {
-		glog.Errorf("cannot add %s to %s ", name, bridge)
-		return err
+	if bridge != "" {
```
gao-feng: In which case BridgeIface is null? the bridge suppose not to be null here.

hypervisor/qemu/network.go
```diff
@@ -33,6 +34,15 @@ func GetTapFd(device, bridge, options string) (int, error) {
 		errno syscall.Errno
 	)
 
+	// check if tapname exists, if true, use existing one instead of create new one
+	if device != "" {
```
gao-feng: device should not be null, and even this device is existed on host, it should be added to bridge too.

btw, use TUNSETIFF of ioctl on tun fd with existed device name will attach the fd to the tap device,
so I think we don't need this change.

hypervisor/network.go
```diff
@@ -201,6 +235,64 @@ func (nc *NetworkContext) removeInterface(id string, result chan api.Result) {
...
+		nc.sandbox.Log(WARNING, "trying update a non-exist interface %s", inf.Id)
+		return fmt.Errorf("interface %q not exists", inf.Id)
+	}
+
+	// only support update some fields: Name, ip addresses, mtu
+	nc.slotLock.Lock()
+	defer nc.slotLock.Unlock()
+
+	if inf.Name != "" {
+		oldInf.NewName = inf.Name
```
gao-feng: Is it necessary to add a new field NewName?Can not save new name to DeviceName?

cli/interface.go
WeiZhang555: Sure. I'll do it.

cli/interface.go
WeiZhang555: Got it

cli/interface.go
WeiZhang555: OK

cli/interface.go
WeiZhang555: Oh, my fault, this should be `context.String("tapname"),`

I modify some logic in this upstream PR, and forgot to update tap setting logic.

cli/network.go
WeiZhang555: I'll keep tracking of this, so far we didn't see any harm

hypervisor/network.go
WeiZhang555: Sure

hypervisor/qemu/network.go
WeiZhang555: I'll change it back.

Because I don't know how to use TUNSETIFF, I choose a more direct way.  But it's ok to me to remove this part of logic.

hypervisor/network/network_linux.go
WeiZhang555: From previous logic:
```
201     if bridge == "" {
202         bridge = BridgeIface
203     }
```

if we pass an empty `bridge`, it will fall down to `BridgeIface`, in my test, it's still empty which means BridgeIface is empty. 

I didn't look deeply into this, in my environment, I can see there's one new `runv0` bridge, and no `hyper0` bridge, not sure if this is related.

hypervisor/network.go
WeiZhang555: There's one trouble here, I add a `NewName` for modifying device name in Guest OS.

in struct `InterfaceCreated`, there are 3 fields in my codes:
```
HostDevice string
DeviceName string
NewName    string
```

`HostDevice` is for tap name

In GuestOS, if we want to add a new interface named `eth5`, and there are two existing network interface named `eth0` and `eth3`, when we insert a new tap device, the new added device will be named `eth1` in Guest, so we should use "eth1" as `DeviceName` which will be used by hyperstart to search for new added network interface.

then we user `eth5` as `NewName`, hyperstart will rename `eth1` to `eth5`. So in guest, we have 3 nic now: `eth0`, `eth3` and `eth5`, next time, newly added tap device will count from `eth1` again.

That's why I need to add a `NewName` field.

In above example, `HostDevice` is tap name, `DeviceName` is `eth1` and `NewName` is `eth5`. I can't remove `NewName` and just set `DeviceName=eth5`, we still need to know the `eth1` name.

That's why it's so complicated and confusing .

WeiZhang555:
ping @gao-feng Updated.

gao-feng:
let's merge it now, thanks for your work!

PR #609, opened 2017-09-28
fix route device name

bergwolf:
After 37799211 ("bugfix & enhancement: add more nic with new name"),
inf.NewName is the true interface name inside of guest.

cc @WeiZhang555


WeiZhang555:
I think you are right, though I didn't test it. :+1:

gao-feng:
LGTM

PR #614, opened 2017-10-11
docs: add more containerd info

Ace-Tang:
add more containerd infomations in docs, make
users more easily to get runv worked with containerd

Signed-off-by: Ace-Tang <aceapril@126.com>

I think this may get more clearly about get runv work with containerd, thanks


gnawux:
Thanks for the contribution @Ace-Tang

PR #621, opened 2017-10-24
Detect the correct irpchip for arm platforms while using kvmtool

Weichen81:
kvmtool supports 3 kinds of virtual irqchip for arm platforms: gicv2, gicv3 and
gicv3-its. But these 3 kinds of virtual irqchip must have correct real irqchip
to support them.  In current code, we fix the irqchip to gicv3. This may work fine
for those systems that are using gicv3 or gicv4. But on a system with gicv2 or gicv2m,
the kvmtool could not start properly.

In this PR, we detect the irqchip type from /proc/interrupt to solve this issue.

### DIFF
--- a/hypervisor/kvmtool/lkvm_arm64.go
+++ b/hypervisor/kvmtool/lkvm_arm64.go
@@ -2,6 +2,40 @@
 
 package kvmtool
 
+import (
+	"bufio"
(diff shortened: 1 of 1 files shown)


gao-feng:
LGTM,thanks!

PR #624, opened 2017-10-25
remove watch-vm-console option of proxy

gao-feng:
hypervisor.VmAssociate and VmContext.Launch both call
watchVmConsole to watch the console of vm, there is no
need to watch again in proxy.

This also fix the fail to connect console problem when
using kvmtool and xenpv driver.

### DIFF
--- a/cli/proxy.go
+++ b/cli/proxy.go
@@ -7,14 +7,12 @@ import (
 	"net"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"syscall"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


laijs:
When proxy is used, hypervisor.VmAssociate() is expected NOT to watch the vm console by design when the time proxy was introduce. hypervisor.VmAssociate() is complicated, there might be some thing wrong.

gao-feng:
issue https://github.com/hyperhq/runv/issues/625

laijs:
the issue is being addressing in #633

PR #628, opened 2017-10-25
hyperstart: gRPC protocol renames

sameo:
- Rename part the execution API
- Add a marshalled config file to the sandbox creation request.

TODO: 

- [X] Modify callers for these API changes.
- [X] Implement gRPC <-> OCI conversion routines
- [X] Unit test the conversion routines
- [X] Unit test the structure copying API


sameo:
@gnawux @bergwolf @sboeuf @mcastelino PTAL

bergwolf:
Is the new field in sandbox creation request required by clear container or is it added just for compatibility with current hyperstart protocol?

cc @laijs @gao-feng

sameo:
@bergwolf The config field in the sandbox is needed for the new clear containers agent, that is based on libcontainer. It needs to get an OCI config file to set the container inside the VM.

bergwolf:
@sameo The StartSandbox() operation is supposed to set pod level configurations (hostname and dns) inside the container, while AddContainer() is the one adding new containers to the sandbox. If libcontainer needs the OCI spec details, how about passing it through AddContainer() instead? The idea is to separate sandbox creation and container creation.

It seems libcontainer also has its own config and converts OCI spec to it in specconv.CreateLibcontainerConfig(). Is it the one requiring OCI spec in the new agent?

sameo:
@bergwolf 
> how about passing it through AddContainer() instead?

Yes, that makes sense. I updated the PR.

> It seems libcontainer also has its own config and converts OCI spec to it in specconv.CreateLibcontainerConfig(). Is it the one requiring OCI spec in the new agent?

Yes.

hyperstart/api/grpc/hyperstart.proto
```diff
@@ -5,15 +5,17 @@ package grpc;
...
 // unstable
 service HyperstartService {
 	// execution
-	rpc AddContainer(AddContainerRequest) returns (google.protobuf.Empty);
-	rpc AddProcess(AddProcessRequest) returns (google.protobuf.Empty);
-	rpc SignalProcess(SignalProcessRequest) returns (google.protobuf.Empty);
-	rpc WaitProcess(WaitProcessRequest) returns (WaitProcessResponse); // wait & reap like waitpid(2)
+	rpc CreateContainer(CreateContainerRequest) returns (google.protobuf.Empty);
+	rpc StartContainer(StartContainerRequest) returns (google.protobuf.Empty);
+	rpc Exec(ExecRequest) returns (google.protobuf.Empty);
```
bergwolf: We should be able to combine StartContainer and Exec just like above AddProcess does. Is there any special requirement to separate starting container and exec?

sboeuf: Yes we can merge both `StartContainer()` and `Exec()` into `StartProcess()` since that's what this is. We have `CreateContainer()` which really means that we are creating a container but without starting any process in it. And we could have `StartProcess()` having a flag to make the difference between the container process or any other process that we want to start.

But at the same time, I think it might be okay to have separate functions for each case since there are 2 distincts use cases. The question is, do we mind having some kind of overlapping behaviors ? I think it's okay, the agent implementation will rely on the same code for the main part of those 2 functions.

sameo: @bergwolf From a code perspective, they will look very similar, but from an API perspective they have different semantics (Their arguments are even different, you don't need to specify a Process when starting an already created container). So I think the API is clearer and more explicit with those 2 separated.

bergwolf: I tend to think we should keep the API list short. But I can live with this. @laijs @gnawux @gao-feng what do you guys think?

hyperstart/api/grpc/hyperstart.proto
```diff
@@ -30,34 +32,42 @@ service HyperstartService {
...
 }
 
-message AddContainerRequest {
+message CreateContainerRequest {
 	Container container = 1;
 	Process init = 2;
+	// config is a marshalled OCI configuration JSON file (a.k.a. config.json).
+	// This is optional and can be used by agents implementing the hyperstart protocol to apply
+	// any resource or security constraints defined by this configuration.
+	string config = 3;
```
laijs: I prefer defining the desired fields in the message type Container&Process.
I don't want an embedded string with another language(json).
We might not use the OCI configuration JSON in the hyperstart since we will use libcontainer directly.

hyperstart/api/grpc/hyperstart.proto
```diff
@@ -30,34 +32,42 @@ service HyperstartService {
...
 }
 
-message AddProcessRequest {
-	string container = 1;
+message StartContainerRequest {
+	string container_id = 1;
+}
```
laijs: I agree on separating AddContainer() into CreateContainer and StartContainer().

hyperstart/api/grpc/hyperstart.proto
```diff
@@ -5,15 +5,17 @@ package grpc;
...
 	// execution
-	rpc AddContainer(AddContainerRequest) returns (google.protobuf.Empty);
-	rpc AddProcess(AddProcessRequest) returns (google.protobuf.Empty);
-	rpc SignalProcess(SignalProcessRequest) returns (google.protobuf.Empty);
-	rpc WaitProcess(WaitProcessRequest) returns (WaitProcessResponse); // wait & reap like waitpid(2)
+	rpc CreateContainer(CreateContainerRequest) returns (google.protobuf.Empty);
+	rpc StartContainer(StartContainerRequest) returns (google.protobuf.Empty);
+	rpc Exec(ExecRequest) returns (google.protobuf.Empty);
+	rpc Signal(SignalRequest) returns (google.protobuf.Empty);
+	rpc Wait(WaitRequest) returns (WaitResponse); // wait & reap like waitpid(2)
```
laijs: The names with "Process" would be clearer.
How about ExecProces()(renamed from AddProcess()) SignalProcess() WaitProcess()(unchanged)?

hyperstart/api/grpc/hyperstart.proto
bergwolf: How about simply putting the OCI specs.Spec and specs.Process in place (with named fields instead of dummy string)? libcontainer in the new agent can use them directly and we do not have to modify hyperstart protocol every time when we want to add new supported fields. Of course it would add a dependency on OCI spec versions but I think it is OK since we want to be OCI compliant anyway.

sameo: @laijs @bergwolf Defining or copying the right field would be ideal, but that means copying most of the OCI `specs.Specs` in this file as there is no OCI specification gRPC proto file.
If that is what you want, I'll create an oci.proto separate file to not pollute the hyperstart.proto one.

hyperstart/api/grpc/hyperstart.proto
sameo: @laijs Fine with me. Done.

hyperstart/api/grpc/hyperstart.proto
laijs: oci.proto sounds good. But it should be along with a verification code(or tool) to prove that the oci.proto and the imported OCI specs are fully matched.

sameo: @laijs @bergwolf I'm implementing `oci.proto` and because of gRPC Go conversion assumptions, we will have to handle 2 different structures:

The gRPC generated Go structure:
```Go
type Spec struct {
	Version     string            `protobuf:"bytes,1,opt,name=Version" json:"Version,omitempty"`
	Process     *OCIProcess       `protobuf:"bytes,2,opt,name=Process" json:"Process,omitempty"`
	Root        *Root             `protobuf:"bytes,3,opt,name=Root" json:"Root,omitempty"`
	Hostname    string            `protobuf:"bytes,4,opt,name=Hostname" json:"Hostname,omitempty"`
	Mounts      []*Mount          `protobuf:"bytes,5,rep,name=Mounts" json:"Mounts,omitempty"`
	Hooks       *Hooks            `protobuf:"bytes,6,opt,name=Hooks" json:"Hooks,omitempty"`
	Annotations map[string]string `protobuf:"bytes,7,rep,name=Annotations" json:"Annotations,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
	Linux       *Linux            `protobuf:"bytes,8,opt,name=Linux" json:"Linux,omitempty"`
}
```

The OCI Spec Go structure:
```Go
type Spec struct {
	Version string `json:"ociVersion"`
	Process *Process `json:"process,omitempty"`
	Root Root `json:"root"`
	Hostname string `json:"hostname,omitempty"`
	Mounts []Mount `json:"mounts,omitempty"`
	Hooks *Hooks `json:"hooks,omitempty" platform:"linux,solaris"`
	Annotations map[string]string `json:"annotations,omitempty"`
	Linux *Linux `json:"linux,omitempty" platform:"linux"`
	Solaris *Solaris `json:"solaris,omitempty" platform:"solaris"`
	Windows *Windows `json:"windows,omitempty" platform:"windows"`
}
```

As you can see those are different structures and we will have to go through the following steps:

```
runtime (host) -> OCI spec structure  ---- conversion ----> gRPC spec structure -> gRPC binary -> gRPC structure --- conversion ----> OCI spec structure -> agent (guest)  
```

So by not tunnelling a marshalled OCI spec through gRPC, we're adding 2 additional conversion layers: from OCI to gRPC (on the host) and then from gRPC to OCI (on the guest).

Are we all ok with that ?

sameo: cc @jodh-intel

sameo: cc @laijs @bergwolf Are we all good with the above approach?

bergwolf: @sameo can you please share your oci.proto? I did some preliminary test and grpc-go did generate expected spec struct w.r.t. the `Solaris *Solaris` field. So I want to understand the issue first.

PS, my test is [here](https://gist.github.com/bergwolf/c1dab2956a422dac4831d5dd9e3c3587)

sameo: @bergwolf Here is mine: https://gist.github.com/sameo/d8dc8c8720a67ae92100fc6d75f78987

For all non scalar fields, `gRPC` will generate a pointer to the struct. So if you create the `Mount` message in your test, you will see that `Spec` becomes:
```Go
type Spec struct {
	Version     string            `protobuf:"bytes,1,opt,name=Version" json:"Version,omitempty"`
	Process     *OCIProcess       `protobuf:"bytes,2,opt,name=Process" json:"Process,omitempty"`
	Root        *Root             `protobuf:"bytes,3,opt,name=Root" json:"Root,omitempty"`
	Hostname    string            `protobuf:"bytes,4,opt,name=Hostname" json:"Hostname,omitempty"`
	Mounts      []*Mount          `protobuf:"bytes,5,rep,name=Mounts" json:"Mounts,omitempty"`
	Hooks       *Hooks            `protobuf:"bytes,6,opt,name=Hooks" json:"Hooks,omitempty"`
	Annotations map[string]string `protobuf:"bytes,7,rep,name=Annotations" json:"Annotations,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
	Linux       *Linux            `protobuf:"bytes,8,opt,name=Linux" json:"Linux,omitempty"`
}
```

and for example you get `Mounts      []*Mount` as opposed to `Mounts      []Mount` for the OCI spec.

sameo: @laijs @bergwolf What about something more structured than my initial proposal:

```
diff --git a/hyperstart/api/grpc/hyperstart.proto b/hyperstart/api/grpc/hyperstart.proto
index 7a062ad..1691d14 100644
--- a/hyperstart/api/grpc/hyperstart.proto
+++ b/hyperstart/api/grpc/hyperstart.proto
@@ -35,10 +35,21 @@ service HyperstartService {
 message CreateContainerRequest {
        Container container = 1;
        Process init = 2;
-       // config is a marshalled OCI configuration JSON file (a.k.a. config.json).
-       // This is optional and can be used by agents implementing the hyperstart protocol to apply
-       // any resource or security constraints defined by this configuration.
-       string config = 3;
+
+       // spec is the container specification.
+       // With OCI its payload will be a marshalled config.json.
+       Spec spec = 3;
+}
+
+message Spec {
+       // type is the specification type. For now we only support the "OCI" type.
+       string type = 1;
+
+       // version is the specification version
+       string version = 2;
+
+       // payload is the marshalled specification payload, e.g. config.json.
+       string payload = 3;
 }
 
 message StartContainerRequest {
```

sameo:
@laijs @gnawux @bergwolf Initial `oci.proto` added.

hyperstart/api/grpc/hyperstart.proto
```diff
@@ -30,34 +35,39 @@ service HyperstartService {
...
+}
+
+message StartContainerRequest {
+	string container_id = 1;
 }
 
-message AddProcessRequest {
-	string container = 1;
+message ExecProcessRequest {
+	string container_id = 1;
```
sboeuf: Just a thought about the syntax. Why don't we use `containerID` here instead of `container_id` ?

hyperstart/api/grpc/hyperstart.proto
```diff
@@ -30,34 +35,39 @@ service HyperstartService {
 	rpc OnlineCPUMem(OnlineCPUMemRequest) returns (google.protobuf.Empty);
 }
 
-message AddContainerRequest {
+message CreateContainerRequest {
 	Container container = 1;
 	Process init = 2;
+	Spec OCI = 3;
```
sboeuf: I agree we need this in order to give all the information related to the container.
However, I am not very familiar with the gRPC code/protocol and I was wondering why we need to specify this Spec in details ? I mean we could use an array of bytes (basically a string) as the third parameter here. From the runtime, we marshal whatever Spec structure we want and from the agent, we unmarshal based on the same structure. I am mentioning this because having the entire spec described here means that we would have to change the protocol if we wanted a new field in the spec.

sameo:
@gnawux @laijs Now with the reflection based conversion routines. I have an initial unit tests proving those routines work as expected. PTAL

sameo:
cc @markdryan

hyperstart/api/grpc/utils.go
```diff
@@ -0,0 +1,140 @@
...
+	if to.Kind() != reflect.Struct && from.Kind() != reflect.Struct {
+		return fmt.Errorf("Can only copy structs into structs")
+	}
+
+	if to.NumField() != from.NumField() {
+		return fmt.Errorf("Structures must have the same number of fields")
+	}
+
+	for i := 0; i < to.NumField(); i++ {
+		if err := copyValue(to.Field(i), from.Field(i)); err != nil {
```
laijs: I think it should be copied based on the filed name instead of the index.

hyperstart/api/grpc/utils.go
```diff
@@ -0,0 +1,140 @@
+package grpc
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
```
laijs: It is very nice to provide this utils. It is very useful.

hyperstart/api/grpc/utils.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if toVal.Kind() != reflect.Ptr || toVal.Elem().Kind() != reflect.Struct ||
+		fromVal.Kind() != reflect.Ptr || fromVal.Elem().Kind() != reflect.Struct {
+		return fmt.Errorf("Arguments must be pointers to structures")
+	}
+
+	toVal = toVal.Elem()
+	fromVal = fromVal.Elem()
+
+	if toVal.NumField() != fromVal.NumField() {
```
markdryan: This check isn't really needed as it is performed again in copyStructValue.

hyperstart/api/grpc/utils.go
```diff
@@ -0,0 +1,140 @@
...
+			return copySliceValue(to, from)
+		case reflect.Map:
+			return copyMapValue(to, from)
+		default:
+			// We now are copying non pointers scalar.
+			// This is the leaf of the recursion.
+			if from.Type() != to.Type() && from.Type().ConvertibleTo(to.Type()) {
+				to.Set(from.Convert(to.Type()))
+				return nil
+			} else {
```
markdryan: The case where the types a different and not convertible is not really handled.  We'll fall into this if statement and presumably panic.  Is this an error case we want to handle explicitly or is the intention to trap the panic at a higher level?

hyperstart/api/grpc/utils.go
```diff
@@ -0,0 +1,140 @@
...
+	if to.Kind() != reflect.Map && from.Kind() != reflect.Map {
+		return fmt.Errorf("Can only copy maps into maps")
+	}
+
+	mapLen := from.Len()
+	to.Set(reflect.MakeMap(to.Type()))
+
+	for j := 0; j < mapLen; j++ {
+		if err := copyValue(to.Index(j), from.Index(j)); err != nil {
+			return err
```
markdryan: I think this code will panic. Value.Index(int) cannot be used on maps.  There are MapIndex and SetMapIndex functions that can be used instead.

hyperstart/api/grpc/utils.go
```diff
@@ -0,0 +1,140 @@
...
+		// Here the destination is not a pointer.
+		// Let's check what's the origin.
+		if fromKind == reflect.Ptr {
+			return copyValue(to, from.Elem())
+		}
+
+		switch toKind {
+		case reflect.Struct:
+			return copyStructValue(to, from)
+		case reflect.Slice:
```
markdryan: Do you need to handle arrays as well?  They have a different reflection type.

hyperstart/api/grpc/utils.go
```diff
@@ -0,0 +1,140 @@
...
+		if err := copyValue(to.Field(i), from.Field(i)); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func copyStruct(to interface{}, from interface{}) error {
+	toVal := reflect.ValueOf(to)
```
markdryan: You may want to consider trapping panics and returning an error here just in case there's an error case that isn't explicitly handled by the code.

sameo:
Unit test added and I also addressed @markdryan comments.

hyperstart/api/grpc/utils.go
sameo: @markdryan We do not need to support arrays for now.

hyperstart/api/grpc/hyperstart.proto
sameo: Because `protoc` translate `foo_id` into `fooID` when being told to compile a .proto file to Go.
See hyperstart.pb.go in this PR.

sboeuf: Ok thanks

hyperstart/api/grpc/utils.go
sameo: @laijs I thought about that option but I'd rather avoid it due to this documented bug:
https://golang.org/pkg/reflect/#pkg-note-BUG

sameo:
@gnawux @bergwolf @laijs This is ready for review now.

hyperstart/api/grpc/utils.go
laijs: Since they can be decoded from json string, which conversion is based on the filed names/tags. I think the problem can be resolved here. (https://github.com/golang/go/blob/a3e013b0824ba53168b5d91abdb6ce191510a89d/src/encoding/json/encode.go#L1231)

I'm worried when the oci spec is changed in future, how would we maintain the compatibilities. (vs. older gRPC, vs. older OCI, and vs. newer OCI)? I seams that copying manually field by field is good. We can use this code, because we can change it to "manually copying" any time when any problem happen, or to a new method any time when we found one.

sameo: @laijs I'm not sure if you're asking for a change here or not. Implementing something similar to the JSON implementation means we would have to maintain additional metadata for all our fields, sort them through a similar function as well, etc...I believe this is an overkill.
I can add an additional check to verify that we're copying fields with the same name and error out when/if we're not. Any OCI config incompatibility would then be caught at unit testing time.

sameo: @laijs I changed the logic in `copyStructValue` to verify that:
* We copy from fields with the same name
* If fields don't have the same name, then we go through a loop searching for the right field in the from structure.

sboeuf: @laijs I think having this full OCI specification structure will be a pain to maintain. Don't you think that having only a `string` field would be easier ? I mean we could pass the entire config.json file serialized as a JSON and we would unmarshal from the agent. That way, we just need to make sure that both runtime and agent rely on the same OCI spec version. I don't see any benefits in having this described into the gRPC protocol description here.

hyperstart/libhyperstart/hyperstart.go
```diff
@@ -25,9 +25,10 @@ type Hyperstart interface {
 	Unpause() error
 
 	APIVersion() (uint32, error)
-	NewContainer(c *hyperstartapi.Container) error
+	CreateContainer(c *hyperstartapi.Container) error
+	StartContainer(container string) error
 	RestoreContainer(c *hyperstartapi.Container) error
-	AddProcess(container string, p *hyperstartapi.Process) error
+	ExecProcess(container string, p *hyperstartapi.Process) error
```
laijs: the hyperstartapi here is actually hyperstart json + serial API.
We are going to covert this interface to use hyperstart grpc API directly.
But it needs more changes and needs to be suspended until the grpc API become a little more mature.
This is just for your information.

hyperstart/api/grpc/oci.proto
```diff
@@ -0,0 +1,456 @@
...
+	// Minor is the device's minor number.
+	int64 Minor = 2;
+}
+
+message LinuxWeightDevice {
+	// Major is the device's major number.
+	int64 Major = 1;
+
+	// Minor is the device's minor number.
+	int64 Minor = 2;
```
laijs: The test doesn't test the copying about it, which is required and important since LinuxWeightDevice in oci.Specs embeds linuxBlockIODevice as an anonymous field here.

sameo: Good point. I fixed it now and the code supports copying embedded structures.

hyperstart/api/grpc/oci.proto
```diff
@@ -0,0 +1,456 @@
...
+	uint64 RealtimePeriod = 5;
+
+	// CPUs to use within the cpuset. Default is to use any CPU available.
+	string Cpus = 6;
+
+	// List of memory nodes in the cpuset. Default is to use any available memory node.
+	string Mems = 7;
+}
+
+message linuxBlockIODevice {
```
jodh-intel: This can be dropped I think as the members are now part of `LinuxWeightDevice` and `LinuxThrottleDevice`.

hyperstart/api/grpc/oci.proto
```diff
@@ -0,0 +1,456 @@
...
+
+message Spec {
+	// Version of the Open Container Initiative Runtime Specification with which the bundle complies.
+	string Version = 1;
+
+	// Process configures the container process.
+	OCIProcess Process = 2;
+
+	// Root configures the container's root filesystem.
+	Root Root = 3;
```
jodh-intel: Shouldn't this (and all the other optional entries like `ConsoleSize`, `Mounts`, ...) be annotated with `optional` to comply as closely as possible with the OCI spec?

hyperstart/api/grpc/hyperstart.proto
jodh-intel: It might be worth pointing out here that this representation implies that the agent process **must** be updated "in lock step" with (at the same time as) the other system gRPC elements. For example, if `oci.proto` is updated to support a new version of the OCI spec (and `Spec.Version` is increased) that will clearly alter the format of the messages meaning the agent won't be able to unpack them unless it too has been updated to handle `OCI.Version`. It won't know that though as, ironically, it won't be able to unpack the `OCI` message to query `Version` and check if it can handle that.

It also potentially implies that the agent might need to support older versions of the spec which don't actually change the representation of `Spec` (but might for example have certain fields that are mandatory whereas a newer version of the spec makes them optional).

The above has an impact on system upgrade: it implies that if the gRPC format changes, all containers need to be stopped (or restarted) to pick up the new gRPC message formats (and associated assets including the new agent).

hyperstart/api/grpc/oci.proto
sameo: @jodh-intel The latest protobuf version (v3) no longer supports `optional`:
https://developers.google.com/protocol-buffers/docs/reference/go-generated#fields

That's unfortunate, but seems it's been a long debated design decision among the protobuf community members.

hyperstart/api/grpc/oci.proto
sameo: ah, good call. I will update `oci.proto`

sameo:
@jodh-intel 
> It might be worth pointing out here that this representation implies that the agent process must be updated "in lock step" with (at the same time as) the other system gRPC elements. For example, if oci.proto is updated to support a new version of the OCI spec (and Spec.Version is increased) that will clearly alter the format of the messages meaning the agent won't be able to unpack them unless it too has been updated to handle OCI.Version. It won't know that though as, ironically, it won't be able to unpack the OCI message to query Version and check if it can handle that.

Yes, that's correct. A few comments:

* We could somehow mitigate that by adding a redundant `Version` field to an `OCISpec` message that would contain the version and the `Spec` itself. At least the agent would know which version of the spec it receives and could act upon it or return a meaningful warning (I don't think it should error out when failing to unpack the gRPC OCI spec message).
* The OCI specs are backward compatible and follow the semver 2.0.0 semantics. So we are guaranteed (famous last words) that we would have to do the lockstep changes *only* when moving to a new major OCI spec version. In which case I think the overall project should also move to a new major version as well. I don't think it's reasonable to claim that e.g. Clear Containers 3.x will support both OCI 1.x and 2.x.

hyperstart/api/grpc/oci.proto
sameo: @jodh-intel Done

hyperstart/api/grpc/oci.proto
jodh-intel: Ah - thanks. I wasn't aware of that.

jodh-intel:
I've just found the `Any` message type which looks like it might offer an alternative way to represent the OCI spec:

- https://developers.google.com/protocol-buffers/docs/proto3#any

sameo:
@laijs Do you have any other concerns/comments on this PR?

gnawux:
I think it in a good shape now, we could push it forward.

sboeuf:
Why is this still not merged ?

laijs:
could you remove the last grpc stream commit, and I merge it? @sameo 
let grpc stream go with a new pr.

laijs:
@sboeuf I felt the API not completed. Example, it seems the storage APIs still missing.
The other part can go within new PRs. I will merge this one.

sameo:
@laijs Streams commit removed.

gnawux:
@laijs @sameo @bergwolf @sboeuf @jodh-intel  Based on our discussion on skype, I will send pr of the protocol to kata-containers/runtimes repo, which will help us accelerating the following discussion and development. I will take a snapshot of the protocol in this PR, and we could move our discussion to the correspond new PR.

sameo:
@gnawux thx

PR #633, opened 2017-11-14
proxy: Wait READY message for before send messages to hyperstart

Weichen81:
Hi,

After the runv update to the latest upstream. I found the runv could not work properly with the lkvm.
There are two reasons cause such result.
1. We have moved the VM console connect code to a standalone function. But we call the function
    before grpcSock created. But this function could not return before proxy.NewServer completed.
2. We have ignore the READY message. This will cause the communication between runv and hyperstart
    could not work properly, especially on some old server, the VM could not boot very quickly.

Regards,
Wei Chen

### DIFF
--- a/cli/proxy.go
+++ b/cli/proxy.go
@@ -45,8 +45,9 @@ var proxyCommand = cli.Command{
 			context.String("hyperstart-stream-sock") == "" || context.String("proxy-hyperstart") == "" {
 			return err
 		}
+
 		glog.Infof("agent.NewJsonBasedHyperstart")
-		h, _ := agent.NewJsonBasedHyperstart(context.String("vmid"), context.String("hyperstart-ctl-sock"), context.String("hyperstart-stream-sock"), 1, false, false)
+		h, _ := agent.NewJsonBasedHyperstart(context.String("vmid"), context.String("hyperstart-ctl-sock"), context.String("hyperstart-stream-sock"), 1, true, false)
 
 		var s *grpc.Server
 		grpcSock := context.String("proxy-hyperstart")


gao-feng:
@laijs 
please help to review proxy related pr, thanks!

hypervisor/kvmtool/lkvm.go
```diff
@@ -439,6 +439,10 @@ func (kc *KvmtoolContext) Save(ctx *hypervisor.VmContext, path string) error {
 }
 
 func (kc *KvmtoolContext) ConnectConsole(console chan<- string) error {
+	if kc.conPty == "" {
```
gao-feng: This check if ok, and we should implement Save interface for kvmtool, so the conPty will not be nil.

Weichen81: While reading this code, I have some doubt about calling watchConsole in Launch and VmAssociate.
I am not sure why we have to open VM console device twice?

gao-feng: I don't know if I get your point correctly, is this pr https://github.com/hyperhq/runv/pull/624 relate to your doubt?

Weichen81: Yes, that PR#624 is relate to my doubt. Thanks Feng!

gao-feng:
lkvm related commits looks good to me.
@laijs  please help to review proxy related commit, thanks

cli/proxy.go
```diff
@@ -106,6 +98,15 @@ var proxyCommand = cli.Command{
...
 
+		if context.String("watch-vm-console") != "" {
+			glog.Infof("watchConsole() sock: %s", context.String("watch-vm-console"))
+			err = watchConsole(context.String("watch-vm-console"))
+			if err != nil {
+				glog.Errorf("watchConsole() failed, err: %#v", err)
+				return err
+			}
+		}
+
```
laijs: I don't understand "Without proxy serve grpc, the watchConsole will be blocked
at utils.UnixSocketConnect(console)." 
What's the dependence behind it?
And the code is moved after "s.Serve(l)", which means the proxy is done. We don't want to watch the console after the proxy is done.

The reason for the blocking at utils.UnixSocketConnect(console) might be the proxy didn't wait the READY message. If so, it is fixed by the first commit in this PR.

Weichen81: @laijs I have checked code again. The correct change would be removing the " watchConsole(context.String("watch-vm-console"))". Because in lkvm driver the console socket
would not be created any more. Current watch vm console has used a channel to log VM message. This UnixSocketConnect will be blocked here forever. Move the watchConsole behind the "s.Serve(l)" just make watchConsole do not block the grpcSock creating. It just covers up the real issue. I will send another patch to remove the watchConsole from createProxy.

>I1117 10:54:54.422648   22017 json.go:194] SB[vm-ALZDANbZzu] hyperstart API version:4244, VM hyperstart API version: 4244
E1117 10:55:03.704558   22017 proxy.go:74] watchConsole() failed, err: &net.OpError{Op:"dial", Net:"unix", Source:net.Addr(nil), Addr:(*net.UnixAddr)(0x442024f1e0), Err:(*os.SyscallError)(0x442024f200)}

The function "newHyperstart" will wait grpcSock for 10 seconds.
		for i := 0; i < 500; i++ {
			if _, err := os.Stat(grpcSock); !os.IsNotExist(err) {
				break
			}
			time.Sleep(20 * time.Millisecond)
		}
But, from above log, you can see that, the watchConsole block the grpcSock create for 10 seconds.

Weichen81: @laijs 
Maybe add some check code into watchConsole function is better than removing it.
Because maybe some hypervisors still use unix sock to watch VM console message:
>	_, err := os.Stat(console)
	if err != nil {
		if os.IsNotExist(err) {
			glog.Infof("Console sock %s isn't existed, using channel", console)
			return nil
		}
		return fmt.Errorf("%s existed, someone may be in service", console)
	}

laijs: I think:
First, try to find the possible way to connect to the console for lkvm.
If there is none, the way could be not to pass the --watch-vm-console argument
to proxy when lkvm driver.

Weichen81: That's make sense. I would do it. Thanks!

Weichen81:
@laijs 
Hi JiangShan, I have updated the change, please give some comments : )

Weichen81:
Hi,

Sorry for disturbing, but any comments?

laijs:
it is some different between lkvm & qemu.
I'm going remove the watch code and add it back later with file-type-detect or other solution.

laijs:
Hello, it's fix by #640

Weichen81:
@laijs 
Thanks, I will rebase this PR to #640

Weichen81:
Hi Guys,

Could you help me to check the Travis CI issue? I am not sure whether this CI issue is caused by this patch.

Regards,
Wei Chen

laijs:
@Weichen81 due to recent changes with unknown reason, shim or proxy fails to receive all the stdio data for the case `runv exec test_busybox command-does-not-exist` sometimes. I restarted the test for you now.
It will be fixed later.

cli/proxy.go
```diff
@@ -45,8 +45,9 @@ var proxyCommand = cli.Command{
 			context.String("hyperstart-stream-sock") == "" || context.String("proxy-hyperstart") == "" {
 			return err
 		}
+
 		glog.Infof("agent.NewJsonBasedHyperstart")
-		h, _ := agent.NewJsonBasedHyperstart(context.String("vmid"), context.String("hyperstart-ctl-sock"), context.String("hyperstart-stream-sock"), 1, false, false)
+		h, _ := agent.NewJsonBasedHyperstart(context.String("vmid"), context.String("hyperstart-ctl-sock"), context.String("hyperstart-stream-sock"), 1, true, false)
```
laijs: thanks for the report.

laijs:
LGTM and the test passed, thanks!

PR #634, opened 2017-11-21
api: grpc: Update network API

sboeuf:
This commit updates gRPC API related to network so that we can get
an hardware address related to the interface. This will help to
determine the right interface inside the VM as we might not know
the name created inside the VM.

Also, it renames UpdateInterface into AddInterfaces, and AddRoute
into AddRoutes for more consistency. Indeed, we can add more than
one route and more than one interface if needed.

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>


sboeuf:
cc @sameo @laijs @bergwolf

laijs:
interfaces are allowed to be updated/removed. So the origin UpdateInterface is chosen.
please add the missing APIs for updating or removing.

sboeuf:
@laijs Yes sure I can add `RemoveInterfaces`, would this be okay ? Don't you think we could add `RemoveRoutes` too ?
Also, the `AddInterfaces` and `AddRoutes` should use `netlink.RouteReplace()`, that way it will add the route if it does not exist and it will update it if it already exists. Does this behavior sounds okay ?

sameo:
@laijs @sboeuf IMHO it would be clearer to have add, delete and update APIs for interfaces.

PR #635, opened 2017-11-22
api: grpc: Add commands to stop and remove a container

sboeuf:
This commit introduces extra commands related to a container, and
specifically to the tear down of a container.

StopContainer() will be used in order to force a container to be
stopped without having to bother finding back the container process.
It will simply kill every processes associated to this container and
finish with the container process.

RemoveContainer() will remove a container from the VM, by tearing down
everything related to this container, so that a new container with the
same ID could be started again later.

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>


sboeuf:
cc @laijs @sameo @gnawux @bergwolf

gnawux:
`SignalProcess` could stop a container, right?

sboeuf:
@gnawux yes we could basically do what `StopContainer` with `SignalProcess`, but there is one difference though. `StopContainer` would wait for the container to be completely stop before to return, while `SignalProcess` will return directly after signalling the process.

gnawux:
Then combo of `SignalProcess` and `WaitProcess`?

sboeuf:
@gnawux well you want the shim to use `WaitProcess` since this is the one who needs to retrieve the exit code. We don't want 2 calls (one from the shim and one from the runtime) onto `WaitProcess` I think.
I agree this is a bit of duplicate but it could be easier to use this one in some cases in order to make things simpler from the runtime. @sameo WDYT ?

@gnawux BTW, are you okay with `RemoveContainer`, I think we need this one, because even if we might be able to do the same implicitly from the agent, it would be really easier to have this action being initiated from the runtime. Moreover, we agreed about the agent being a passive component (protocol wise), and I think this new function fits into this description.

gnawux:
@sboeuf Yes, I am OK with the `RemoveContainer`.

sameo:
@sboeuf @gnawux Also implementing a `StopContainer` semantic with `SignalProcess`+`WaitProcess` would have to be done for each process in the container, while `StopContainer` only takes a container id and let the agent handle the process to container mapping. I think it's a reasonable simplification.

hyperstart/api/grpc/hyperstart.proto
```diff
@@ -55,6 +57,14 @@ message WaitProcessResponse {
 	int32 status = 1;
 }
 
+message StopContainerRequest {
+	string container_id = 1;
+}
```
sameo: You probably want a timeout here as well.

bergwolf: I think StopContainer should force kill the container if agent fails to stop it with the stop signal. Is the timeout used to determine when agent should think stop fails and initiate SIGKILL?

sboeuf: @sameo I am okay with adding a timeout here.
@bergwolf I think the timeout would be about StopContainer returning an error if we could not stop the container. I would use SIGKILL from the beginning with StopContainer.

bergwolf: @sboeuf SIGTERM gives container apps a chance to gracefully stop before being killed by SIGKILL. This is also how `docker stop` works. I think we should follow the behavior if we introduce StopContainer API.

PR #636, opened 2017-11-27
Add new option "cache" to container and volume

teawater:
Signed-off-by: Hui Zhu <teawater@hyper.sh>


hypervisor/qemu/qmp_wrapper.go
```diff
@@ -35,8 +35,8 @@ func defaultRespond(result chan<- hypervisor.VmEvent, callback hypervisor.VmEven
 	}
 }
 
-func newDiskAddSession(ctx *hypervisor.VmContext, qc *QemuContext, filename, format string, id int, readonly bool, result chan<- hypervisor.VmEvent) {
-	args := "drive_add dummy file=" + filename + ",if=none,id=" + "drive" + strconv.Itoa(id) + ",format=" + format + ",cache=writeback"
+func newDiskAddSession(ctx *hypervisor.VmContext, qc *QemuContext, filename, format string, id int, readonly bool, cache string, result chan<- hypervisor.VmEvent) {
+	args := "drive_add dummy file=" + filename + ",if=none,id=" + "drive" + strconv.Itoa(id) + ",format=" + format + ",cache=" + cache
```
laijs: Will it be correct when cache="" ?

gnawux: looks there need a default string or a check on empty

teawater:
Add a new version that checks "cache" before use it.

PR #646, opened 2017-12-13
Fix xen build

gnawux:
#597 introduced a bug, which blocked the xen build of runv and hyperd


gnawux:
passed
@laijs ptal

laijs:
LGTM

PR #667, opened 2018-02-01
qemu: clear persist flag on tap device

liwei:
In the scenario which tap device was not created by qemu but passed
in(e.g, with some cni plugin), there was IFF_PERSIST flag set on it,
the flag needs to be cleared to make sure tap device removed cleanly
on qemu exit.


laijs:
LGTM

PR #671, opened 2018-02-11
enable kata shim

bergwolf:
Make sure kata-shim is installed at `/usr/libexec/kata-containers/kata-shim` and pass `--agent kata` to runv cli. Then runv will invoke kata-shim and use it to proxy for agent.

Like:
```
sudo runv --kernel /var/lib/hyper/kernel --initrd //golang/src/github.com/hyperhq/hyperstart/build/kata-initrd.img --agent kata --debug run test1
```

The last commit needs some extra attentions because it changes runv API such that `vm.AddContainer` actually creates a container in the guest, and therefore it must be called **after** all container resources are properly inserted.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -19,15 +19,14 @@
   version = "v0.11.4"
 
 [[projects]]
+  name = "github.com/containernetworking/plugins"
+  packages = ["pkg/ns"]
+  revision = "dd8ff8a5cf537dcfc705e14a1465f82817869bf9"
+  version = "v0.7.0"
+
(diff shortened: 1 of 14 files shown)


bergwolf:
It's working with kata agent/shim/initramfs image.

```
$sudo runv --kernel /var/lib/hyper/kernel --initrd //golang/src/github.com/hyperhq/hyperstart/build/kata-initrd.img --agent kata run test1
/ # ls
bin       dev       etc       lib       proc      root      sys       testdata  tmp       usr       var
```

PR #673, opened 2018-03-22
when qmp is closed, close ongoing qmp sessions

gnawux:
by closing the result channel


laijs:
LGTM

PR #674, opened 2018-03-22
Fix network performance issue on arm64 for runv

jongwu:
Network performance in runv is lower significantly
than in virtual machine started by qemu,as vnet_hdr
feature is not enabled in runv when get access to
a tun/tap device. Adding CIFF_VNET_HDR flag to GetTapFd
func will fixed that issue.

Signed-off-by: Jianyong Wu <Jianyong.Wu@arm.com>
Signed-off-by: Wei Chen <Wei.Chen@arm.com>

@liwei @gnawux @bergwolf  help check it.


gnawux:
+1
thank you @jongwu , sound great

laijs:
LGTM

PR #679, opened 2018-05-14
launchQemu: Add runvpid to args

teawater:
Put runvpid to "-pidfile" will not affect qemu because qemu will use the second "-pidfile" and ignore the first one.

It will be shown in cmd line of qemu that will help admin to make clear the qemu is current hyperd or the crashed hyperd.


laijs:
Is it possible to use a different argument?

teawater:
OK, I will try another one.

PR #684, opened 2018-07-04
BootConfig: add GDBTCPPort

teawater:
If GDBTCPPort set to non-zero, launchQemu will add "-gdb tcp::xxx" to qemu
arguments.
Then GDB can connect this port to debug guest kernel.

Signed-off-by: Hui Zhu <teawater@hyper.sh>


gnawux:
LGTM

IIUC it will add gdb socket and disable the handshake/watchdog timer

PR #685, opened 2018-07-19
Use host ram size and cpus number as maxmem and maxcpus

teawater:


laijs:
LGTM
please update the libvirt driver too.

PR #690, opened 2019-01-09
kill shim carefully

jimoosciuc:
In some cases, the original code will trigger panic


cli/container.go
```diff
@@ -231,6 +231,7 @@ func addProcess(options runvOptions, vm *hypervisor.Vm, container, process strin
 		}
 	}()
 
+	var shim *os.Process
```
bergwolf: Nice catch!

bergwolf (approved):

bergwolf:
Thanks!

